PR #2, opened 2020-05-19
Creating a glossary with a complete cross-reference graph.

gvwilson:
1.  Reformatting the human-edited glossary (details below).
2.  Creating `bin/build-graph.py` to:
    1.  Read the new-format glossary.
    2.  Extract implicit forward references from Markdown definitions (English only).
    3.  Fill in the list of forward references and construct a list of backward references.
    4.  Check along the way that all used terms are defined.

The new-format glossary is:

```
- slug: some_underscored_slug
  en:
    term: "term in English"
    def: >
      Full definition, possibly including [links](#links) to other terms
      (which then do not have to be linked to in the 'ref' section).
  ref:
    - some_other_slug
```

`bin/build-graph.py` will add keys to the `ref` section (which is *not* nested
under any particular language) and create a `usedby` section listing the slugs
of all the terms that point at this one.

### DIFF
--- a/bin/build-graph.py
+++ b/bin/build-graph.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+
+'''Generate YAML with all cross-references filled in, failing if there are undefined terms.'''
+
+import sys
+import re
+import yaml
+
+
+# Match internal Markdown links.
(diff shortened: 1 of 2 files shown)


elletjies:
[@all-contributors](https://github.com/all-contributors) please add @gvwilson for doc,review,question,translation

PR #4, opened 2020-06-12
Add entries by PR

fmichonneau:
WIP: ultimately, I'd like to leverage the machinery of the `usethis::pr_*` functions family to allow contributors to submit new entries.


ian-flores:
@fmichonneau I think this is more related directly with the R package so I'll close the PR here.

PR #12, opened 2020-06-24
Four use cases

gvwilson:
Four use cases for the glossary packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,55 @@ A glossary entry is structured like this:
     should `define('something')` in RStudio put the definition in the help pane
     (and if so, should it hyperlink to terms that the definition depends on)?
 
+## Use Cases
+
+1.  **Linking to a definition.**
+    1.  *Amari* writes a lesson in R Markdown that introduces some new terms.
+    1.  She has defined the language to be Spanish using the `glossary/language` key
+        in the YAML header,
+        but has not changed any other settings.
+    1.  She adds an inline code block `` `r gdef('linear-model', 'Linear models')` ``
+        to her lesson.
+    1.  When she knits her document,
+        the inline code block produces the HTML
+        `<a href="http://carpentries.org/glossary/es/#linear-model" class="glossary-definition">Linear Models</a>`
+
+1.  **Checking a lesson.**
+    1.  *Beatriz* has made some changes to a lesson she inherited from *Amari*,
+        and wants to check that it is still consistent.
+    1.  She runs a command-line script that:
+        1.  Reads the R Markdown file.
+        1.  Extracts the terms under the `glossary/defines` key.
+        1.  Searches the body of the document for calls to `gdef(...)`.
+        1.  Checks that every term listed in `glossary/defines` is referenced in the document body,
+            and that every term referenced in the document body is mentioned in `glossary/defines`.
+
+1.  **Finding lessons.**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,6 +109,55 @@ A glossary entry is structured like this:
...
+    1.  She has defined the language to be Spanish using the `glossary/language` key
+        in the YAML header,
+        but has not changed any other settings.
+    1.  She adds an inline code block `` `r gdef('linear-model', 'Linear models')` ``
+        to her lesson.
+    1.  When she knits her document,
+        the inline code block produces the HTML
+        `<a href="http://carpentries.org/glossary/es/#linear-model" class="glossary-definition">Linear Models</a>`
+
+1.  **Checking a lesson.**
```
ian-flores, line 125: I strongly feel this shouldn't be a use case and that it instead should be part of the packages.

ian-flores: Sorry, it should be a use case but wrapped under a function.

README.md
```diff
@@ -109,6 +109,55 @@ A glossary entry is structured like this:
...
+
+1.  **Finding lessons.**
+    1.  *Amari* writes a lesson in R Markdown.
+        She adds the `glossary` key to its YAML metadata
+        and indicates that the lesson requires the term `correlation`
+        and defines the term `regression`.
+    1.  *Beatriz* is writing a lesson on linear models.
+        She adds YAML metadata indicating that
+        the lesson requires the term `regression`.
+    1.  To find prerequisite lessons she can recommend to her students,
```
ian-flores, line 143: I don't feel this should be a use case. It feels more natural for it to be a function which points to a folder and then we get all of the Rmds in that folder and run ourselves the rmarkdown function.

ian-flores: Sorry, it should be a use case but wrapped under a function.

README.md
gvwilson: I don't understand - as written, it *is* a function call. Can you give an example of what the user would type in their R Markdown?

README.md
gvwilson: :+1: there should be a function that can be called directly from R with a list of filenames. There should also be a command-line script that calls that function that can be integrated into builds.

README.md
```diff
@@ -109,6 +109,55 @@ A glossary entry is structured like this:
     should `define('something')` in RStudio put the definition in the help pane
     (and if so, should it hyperlink to terms that the definition depends on)?
 
+## Use Cases
+
+1.  **Linking to a definition.**
+    1.  *Amari* writes a lesson in R Markdown that introduces some new terms.
+    1.  She has defined the language to be Spanish using the `glossary/language` key
```
zkamvar, line 116: Clarification question: Does `glossary/language` mean that the yaml header will have:

```yaml
glossary:
  - language: es
```

gvwilson: :+1: If there's a cleaner way to write nested YAML keys inline, please let me know - path notation was the only thing I could think of.

README.md
ian-flores: We cleared this up in a conversation/call. As it is written it is now clear.

README.md
ian-flores: Yes @zkamvar that would be the structure. Closing this comment to merge the PR.

PR #15, opened 2020-07-13
Expanding and checking definitions.

gvwilson:
1. Adding more terms to `glossary.yml` to be defined later.
2. Building utils/check-glossary.py to check internal consistency.

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -141,7 +141,7 @@
   en:
     term: "branch"
     def: >
-      See [Git branch](git_branch).
+      See [Git branch](#git_branch).
 
 - slug: branch_per_feature_workflow
   en:
--- a/glossary.yml
+++ b/glossary.yml
@@ -1,16 +1,14 @@
-- slug: plus_one
+- slug: abandonware
   en:
-    term: "+1"
+    term: "abandonware"
     def: >
-      A vote in favor of something.
-  fr:
-    term: "+1"
+      FIXME
+
+- slug: absolute_error
+  en:
+    term: "absolute error"
     def: >
-      Un vote en faveur de quelque chose.
-  es:
-    term: "+1"
-    def: >
-      Un voto a favor de alguna cosa.
+      FIXME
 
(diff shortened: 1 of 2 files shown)


gvwilson:
Please merge now - it'll give people lots to define...

glossary.yml
```diff
@@ -127,6 +158,34 @@
...
+      A collection of files, logs, or related information that adequately
+      describes either an unexpected output of some code or program or an unexpected
+      error or warning. This information is used to help find and fix a bug in the
+      program or code.
+
+- slug: bug_tracker
+  en:
+    term: "bug tracker"
+    def: >
+      A system that tracks and manages [reported bugs](#bug_report) for a software program,
```
zkamvar, line 174: There are two methods for creating cross-links here. `[text](#ref)` and `[text](ref)` (https://github.com/carpentries/glosario/pull/15/files#diff-d85e4245bdeda1fb596b9ff8ade0690fR144). Which one is preferred? If these are on separate pages, I would assume that the `[text](ref)` syntax is preferred

gvwilson: This is a bug in the example - all cross-references should use `[text](#ref)` so that they resolve in-page.

glossary.yml
```diff
@@ -107,16 +131,23 @@
...
   en:
     term: "branch"
     def: >
-      A snapshot of a version of a Git repository. Multiple branches can capture
-      multiple versions of the same repository.
-  ref:
-    - feature_branch
-    - fork
-    - master_branch
+      See [Git branch](git_branch).
```
zkamvar, line 144:
```suggestion
      See [Git branch](#git_branch).
```

glossary.yml
ian-flores: Resolved with [text](#ref) as the standard.

ian-flores:
I'm going to merge taking into consideration the following:

- `FIXME` entries are for contributors
-  Link standard is `[text](#ref)`

PR #37, opened 2020-08-02
Adding logo

gvwilson:
Adding a parrot logo for the project.

1.  Adding logo file to `static`.
2.  Updating CSS to get two-column layout for title.
3.  Adding `_includes/title.html` to display title.
4.  Modifying pages to include titles.


_config.yml
```diff
@@ -19,14 +19,17 @@ languages:
   title: Woordelys
 - key: en
   name: English
+  title: Glossary
```
ian-flores, line 22: Why two title entries per key for these 3 languages?

ian-flores: Spanish and French have the same problem.

gvwilson: That was a merge glitch - fixed.

PR #41, opened 2020-08-03
try to deploy only from main repo

fmichonneau:
don't try to deploy from forks (as it will fail)

### DIFF
--- a/.github/workflows/build-and-deploy.yaml
+++ b/.github/workflows/build-and-deploy.yaml
@@ -21,7 +21,7 @@ jobs:
       run: ls -alh _gh-site && ls -alh _gh-site/_data
 
     - name: Deploy
-      if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+      if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && github.repository == 'carpentries/glosario' }}
       uses: maxheld83/ghpages@v0.3.0
       env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-deploy.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
       run: ls -alh _gh-site && ls -alh _gh-site/_data
 
     - name: Deploy
-      if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+      if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && github.repository == 'carpentries/glosario' }}
```
ian-flores, line 24: Why are we doing this?

fmichonneau:
this action takes care of copying the `glossary.yml` file from the root folder to the `_data` folder (where Jekyll expects it). Then it copies the entire content of the repository into a new folder that gets imported into the `gh-pages` branch. From there, we rely on GitHub pages to generate the site.
 
This last part (pushing to the `gh-pages`) can only be done with the correct PAT. When others work on their forks they would see a failure at this step if we don't add this check because they wouldn't be able to push to the `gh-pages` branch. Does that make sense?

ian-flores:
Ah got it. I wasn't clear on this last part so was confused why we wanted to block it to only this repo, now it makes sense. Thanks @fmichonneau !

PR #48, opened 2020-08-13
Improving glossary checks.

gvwilson:
1.  Add `make check` target to `Makefile` to do high-level glossary consistency check.
1.  Add `make checkall` target to `Makefile` to report missing entries in all languages.
1.  Teach `utils/check-glossary.py` to read `_config.yml` configuration as well as glossary in order to check for unknown languages.


gvwilson:
I think this is ready to review and merge.

PR #50, opened 2020-08-13
lint _config.yml

fmichonneau:
add linting of the `_config.yml` file as part of the "Lint" GitHub Actions


zkamvar:
I just submitted #51, but looking at the GA, I'm wondering if this file actually does anything? How should we proceed with linting?

fmichonneau:
I removed the per-command configuration (and added the extra option provided in #51) and it seems that the `.yamllint.yml` is now being taken into account.

zkamvar:
LGTM! I'll go ahead and rescind #51 since it's included here

PR #51, opened 2020-08-13
update linter to catch incorrect indentation

zkamvar:
This would be able to catch the problems that were fixed in #45 (ideally)

https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.empty_values


zkamvar:
This is now being taken care of in #50

PR #68, opened 2020-08-22
Utility to merge glossary files.

gvwilson:
This will allow other projects to use our glossary and their own
(similarly-formatted) glossary files by merging ours with theirs to produce a
unified files for...well, whatever. To test:

    $ utils/merge.py -w test/words.txt test/alpha.yml test/beta.yml

Built on #67 - please review and merge that one first.


ian-flores:
Not sure who will review this PR (I saw @zkamvar being requested above) but just a note on being careful to NOT merge the `glossary.yml` file. It seems like this PR does a couple of additions but it shouldn't be part of the glossario in my opinion.

zkamvar:
Agree with @ian-flores. Merging is in the domain of the -py and -r APIs, not the main repo.

gvwilson:
I'm confused: there's nothing Python-specific or R-specific about combining two YAML files. See #71 for a use case.

zkamvar:
I may be over-engineering or confused, but, I was working under the assumption that the glosario repo would contain the glossary and the tools to host the website/check the glossary. The -py and -r packages would contain tools for accessing and manipulating that glossary in a consistent manner.

If we include a tool for merging the glossary, it would be better in the python and R because it would allow us to ensure that the tool itself is properly tested and make sure that those who use only R also have access to it.

gvwilson:
Your logic is impeccable; my thought was that the task of merging glossary files is independent of program language, and that if I am an R user, I might not want to have to install glosario-py in order to do that task. I'm happy to relocate the widget if y'all would prefer.

Issue #53, opened 2020-08-15
15/08/2020: Idea for Template - languages other than English


beatrizmilz:
Hi!
I would like to say an idea.
I think it would be good to show in the pages of Glosario of other languages rather than English, the term in English aswell.

A example situation to ilustrate: Gabriela speaks portuguese. She is reading a tutorial about R in english and she wants to know what is a `current working directory`. She opens the Glosario in portuguese ( https://carpentries.github.io/glosario/pt/ ) and tries to find it. But in portuguese this term is `diretório de trabalho` , and does not find the definition. If glosario also showed the term in English (because is the language most used in tutorials, courses, books), it would be easier to find. 

Example: 
![image](https://user-images.githubusercontent.com/42153618/90314707-21dcfb80-deec-11ea-978d-c4f6ba804d73.png)

I'm not sure if this goes out of the scope of Glosario, is just an idea that I think it could make easier to people consult Glosario while studying DS. 

Thank you!

beatrizmilz:
GitHub added the label of bug, but it's just a idea. I'm not able to chage the label

elletjies:
@all-contributors please add @beatrizmilz question

PR #73, opened 2020-08-28
A proposal for project decision-making

gvwilson:
Please add thoughts in the thread below.


GOVERNANCE.md
```diff
@@ -0,0 +1,81 @@
...
+    After 10 minutes or when no one has anything further to add (whichever comes
+    first), the facilitator calls for a yes-or-no vote on the question: "Should
+    we implement this decision over the stated objections?"  If a majority votes
+    "yes" the proposal is implemented.  Otherwise, the proposal is returned to
+    the sponsor for further work.
+
+## Suffrage
+
+1.  The following people are currently voting members of this project:
+    -   [Ian Flores Siaca](https://audassimus.com/)
```
ian-flores, line 56: My page is ianfs.dev

GOVERNANCE.md
```diff
@@ -0,0 +1,81 @@
...
+    -   [Zhian Kamvar](https://zkamvar.netlify.app/)
+    -   [François Michonneau](https://francoismichonneau.net/)
+    -   [Greg Wilson](http://third-bit.com/)
+
+2.  To become a voting member:
+    1.  A current member must sponsor a proposal to add the candidate.
+    2.  That proposal must pass as outlined in the "Making Decisions" section of
+        this page.
+
+3.  Voting members may resign from the project at any time, or lose their vote
```
ian-flores, line 66: Would this be by 50% as well?

GOVERNANCE.md
```diff
@@ -0,0 +1,81 @@
...
+    if a proposal to that effect is passed as outlined in the "Making Decisions"
+    section of this page.
+    1.  A proposal to remove someone's vote may not contain any other business.
+    2.  Each person must be the subject of a separate proposal.
+    3.  The person in question has the right to vote on the proposal to remove
+        their vote.
+
+## Recording Decisions
+
+Proposals that have been accepted are listed below in reverse chronological
```
ian-flores, line 76: Do we want to list them here? How to keep track of it? Manual?

gvwilson: Yes and yes until the group decides on something else, such as a wiki page in the GitHub repo or a separate `RESOLUTIONS.md` file (both of which feel like overkill for now).

GOVERNANCE.md
gvwilson: Yes - usual rules apply.

GOVERNANCE.md
gvwilson: Thanks.

GOVERNANCE.md
```diff
@@ -0,0 +1,81 @@
...
+
+5.  If all of the group likes or can live with the proposal, it is immediately
+    accepted with no further discussion.
+
+6.  If most of the group is uncomfortable with the proposal, it is postponed for
+    further rework by the sponsor.
+
+7.  If any members are uncomfortable they can briefly state their objections.  A
+    timer is then set for a brief discussion moderated by the facilitator.
+    After 10 minutes or when no one has anything further to add (whichever comes
```
zkamvar, line 47: I think this is a good standard for our current group size, but the time limit here does present a barrier for those who are not able to communicate fluently (e.g. those who have disabilities related to verbal communication and/or are not relatively comfortable speaking in English). I would suggest to use a modification from the text for Martha's rules of something like "during roll call, members will state how much time they reasonably require for comments. For example, an abled English speaker may request 2 minutes as reasonable." This way, we can be equitable for any members who require more time to speak.

_includes/glossary.html
```diff
@@ -21,7 +21,7 @@
 {%- for slug in actual -%}
   {% assign item = gloss | where: "slug", slug | first %}
   {% if item[language] %}
-  <dt id="{{item.slug}}">{{item[language].term}}</dt>
+<dt id="{{item.slug}}">{{item[language].term}} {% if language != 'en' %}<span class="in-english">(in English: {{item['en'].term}})</span>{% endif %}</dt>
```
gvwilson, line 24: 1. I don't know why this diff is showing up - `master` has this line and so does the `governance` branch.
2. Why are we giving special status to English?

glossary.yml
```diff
@@ -832,6 +838,11 @@
     def: >
       The folder or directory location that the program is operating in. Any action
       taken by the program occurs relative to this directory.
+  pt:
+    term: "diretório de trabalho"
```
gvwilson, line 842: I don't know why these diffs are showing up - they're in `master` and in `governance`, so they...shouldn't?

GOVERNANCE.md
```diff
@@ -0,0 +1,86 @@
...
+    -   Who can live with the proposal?
+    -   Who is uncomfortable with the proposal?
+
+5.  If all of the group likes or can live with the proposal, it is immediately
+    accepted with no further discussion.
+
+6.  If most of the group is uncomfortable with the proposal, it is postponed for
+    further rework by the sponsor.
+
+7.  If any members are uncomfortable:
```
gvwilson, line 45: Amended as per @zkamvar's comments.

_includes/glossary.html
zkamvar: This line was hoovered in c9aad1005fd4af15a88a5236f221f5d6a45704f7, which is from https://github.com/carpentries/glosario/pull/56 as a potential solution to #53.

GOVERNANCE.md
```diff
@@ -42,13 +42,18 @@ to date.
...
-    first), the facilitator calls for a yes-or-no vote on the question: "Should
-    we implement this decision over the stated objections?"  If a majority votes
-    "yes" the proposal is implemented.  Otherwise, the proposal is returned to
-    the sponsor for further work.
+7.  If any members are uncomfortable:
+    1.  They can briefly state their objections and tell the facilitator how
+        much time they reasonably require to make comments.
+    1.  The facilitator then sets a time limit for discussion at 10 minutes
+        plus whatever upward adjustment needed to meet any special needs
+        participants may have.
```
zkamvar, line 50:
```suggestion
        plus whatever upward adjustment needed to accommodate for any barriers
        to verbal communication.
```

PR #75, opened 2020-09-01
Proposal for governance (updated)

gvwilson:
Fresh pull request without complications - includes suggestions from @zkamvar.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,0 +1,86 @@
+---
+permalink: /governance/
+---
+# Governance
+
+This project uses [Martha's Rules][marthas-rules] for making decisions.  The
+first section below describes these rules; the second lists voting members of
+the project and explains how to become one, while the third lists decisions made
+to date.
+
+[marthas-rules]: https://journals.sagepub.com/doi/10.1177/088610998600100206
+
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -0,0 +1,86 @@
...
+    accepted with no further discussion.
+
+6.  If most of the group is uncomfortable with the proposal, it is postponed for
+    further rework by the sponsor.
+
+7.  If any members are uncomfortable:
+    1.  They can briefly state their objections and tell the facilitator how
+        much time they reasonably require to make comments.
+    1.  The facilitator then sets a time limit for discussion at 10 minutes
+        plus whatever upward adjustment needed to meet any special needs
```
zkamvar, line 49: the phrase 'special needs' is abelist language. I would suggest to change it to "account for any barriers to verbal communication" which refocuses the adjustment on the barriers, not the person.

gvwilson: After some local consultation, I changed it to "plus whatever upward adjustment needed to account for any barriers to participation people may have", since barriers may include things like a poor internet connection.

zkamvar (approved):

PR #77, opened 2020-09-01
Small tool for merging multiple glossary files.

gvwilson:
Use cases:

1. The authors of a book want to combine this glossary with some terms they define themselves. Using `utils/merge-glossaries.py`, they can merge `glosario/glossary.yml` and `book/extra-terms.yml`.

2. The authors of the same book want to maintain their custom glossary file. Using `utils/intersect-glossaries.py`, they can check for duplicated entries and remove them from their custom glossary add-on file.

These tools are written in Python, but that is incidental to their use: they are not intended to be loaded into other programs, but are instead command-line utilities I've needed while working on two book projects.

### DIFF
--- a/utils/intersect-glossaries.py
+++ b/utils/intersect-glossaries.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python
+
+'''Find duplicate entries in two or more glossary files.'''
+
+import sys
+import getopt
+import yaml
+
+
+USAGE = 'intersect glossary [glossary...]'
+
+
+def main():
+    '''Main driver.'''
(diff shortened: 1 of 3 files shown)


zkamvar:
Regarding this tool, I'll leave my comments from the previous PR: https://github.com/carpentries/glosario/pull/68#issuecomment-679210796

This does raise the issue (independent of where this particular tool goes) that we need to decide on what functionality the R and python packages need to implement together. For example, functions to include links to definitions in RMarkdown output is clearly a domain of R and not python, but downloading and storing dictionaries is the domain of both.

PR #78, opened 2020-09-03
More English definitions

gvwilson:
## Author: 

- Greg Wilson

## Language: 

- English

## Terms defined:

- default value
- distro
- Git remote
- keyword arguments
- logging level
- maximum likelihood estimation
- ORCID
- shell
- shell variable
- software distribution
- YAML

## Editor

@zkamvar


glossary.yml
```diff
@@ -2222,8 +2260,8 @@
   en:
     term: "open license"
     def: >
-      A license that permits general re-use, such as the [MIT
-      License][mit_license] or [GPL][gpl].
+      A license that permits general re-use, such as the [MIT License](#mit_license)
+      or [GPL](#gpl).
```
zkamvar, line 2265:
```suggestion
      or [GPL](#gpl) for software and [CC-BY](#cc_by) or [CC-0](#cc_0) for data, prose, or other creative outputs.
```

glossary.yml
```diff
@@ -1738,6 +1757,15 @@
...
       dashes, as in `first-second-third`.
 
+- slug: keyword_arguments
+  ref:
+    - named_argument
+    - variable_arguments
+  en:
+    term: "keyword arguments"
+    def: >
+      Extra arguments given to a function as key-value pairs.
```
zkamvar, line 1767: Are these only a feature of python? If so, it might be useful to mention them and also label that they appear in python help documentation as "kwargs" or "\*\*kwargs".

zkamvar (changes requested):
Overall, looks good :) I'm requesting clarification on keyword arguments.

glossary.yml
gvwilson: Other languages support them as well in various forms.

zkamvar (approved):

zkamvar:
You may merge when ready

PR #84, opened 2020-09-15
Add Contributing Guide

tobyhodges:
At the moment, this includes only guidance for those submitting new terms/definitions. I hope it will be useful to guide first-time contributions e.g. during Hacktoberfest or Carpentries Instructor Training checkout.

If merged, I'm not sure whether this will override the contributing guide currently linked when opening a PR to this repo, which is [the standard Carpentries contibuting.md](https://github.com/carpentries/.github/blob/master/CONTRIBUTING.md).


tobyhodges:
And one question for the reviewer(s): should the second section ("Guide to contributing") be renamed and/or moved to appear above the more specific guidance for adding a new term and definition?

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
...
+please fork [the main Glosario repository][repo]
+and, on a new branch,
+find the term in [`glossary.yml`][glossary] and
+add the two-letter ISO 639 language code such as `en` or `fr`
+as a new top-level key.
+(Refer to the "639-1" column of [this table][iso639-table-en].)
+You can then fill in the `term`, `def`, and (if appropriate) `acronym` for that term in your chosen language.
+See the section above for a full description of the required format for a language term.
+
+## Guide to contributing
```
zkamvar, line 64: I think the Guide to Contributing should come first since it defines the steps people should go through before contributing.

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
+Thank you for your interest in contributing to Glosario!
+We welcome contributions of all kinds.
+If you're here to submit a definition,
+either for a new term
```
zkamvar, line 4: GitHub preserves newline character, so it would be good to keep paragraphs on one line.

CONTRIBUTING.md
```diff
@@ -0,0 +1,96 @@
...
+
+### 2. Adding a definition in another language to an existing term
+
+To add a new definition to an existing term in the glossary,
+please fork [the main Glosario repository][repo]
+and, on a new branch,
+find the term in [`glossary.yml`][glossary] and
+add the two-letter ISO 639 language code such as `en` or `fr`
+as a new top-level key.
+(Refer to the "639-1" column of [this table][iso639-table-en].)
```
zkamvar, line 60: More descriptive for screen readers

```suggestion
(Refer to [the "639-1" column of this table][iso639-table-en].)
```

zkamvar (commented):
It looks good to me with some rearrangement of the sections and tidying up of the text. It would also be good to make sure that the README is up-to-date with our conventions.

fmichonneau:
What do you all think of including an example of a definition with one language and a definition with 2 or more languages directly in this file so contributors could see what it looks like without being overwhelmed by the many definitions are currently in the YAML file?

tobyhodges:
the CRAN example already in there would suffice for a one-language example? I would then add a second example below the section about adding a new definition to an existing term in a different language, showing maybe the [`global_variable` entry](https://github.com/carpentries/glosario/blob/e38b7690848cac9be950873eb6c5c8624188c6f5/glossary.yml#L1474)?

fmichonneau:
> the CRAN example already in there would suffice for a one-language example? 

Sorry I missed that as I didn't notice I was only reviewing the most recent changes. 

> I would then add a second example below the section about adding a new definition to an existing term in a different language, showing maybe the global_variable entry?

That sounds good to me. I don't want to overwhelm future contributors but I think that seeing how multi-language entries are handled would be useful.

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
+Thank you for your interest in contributing to Glosario!
+We welcome contributions of all kinds.
+If you're here to submit a definition,
+either for a new term
+or in another language for an existing term,
+please take a moment to read the guidance below.
+
+## Contribution Workflow
+
+When you're making your contribution(s) to Glosario, please keep the following
```
gvwilson, line 10: end the sentence with "please:" (the rest is redundant)

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+guidance in mind:
+
+1. **Check the [Issues page of the Glosario repository][issues]** to see what terms and definitions other contributors are already working on. This will help you avoid accidentally spending time writing the same term(s)/definition(s) as someone else.
+2. When you know what term or definition you'd like to add, tell others about it by writing **a [new issue] on the main Glosario repository.**
```
gvwilson, line 14: Please specify the language and terms in the title of the issue and/or pull request, e.g., `Romanian: logistic regression`.

gvwilson: Having said that, I suspect people will find this onerous: if I want to add a single term, I'm not going to file the issue and then write the PR - I'm just going to write the PR. Can we ask them to put language/term in the subject of the PR and then also ask contributors to search those as well as issues before contributing? Issues can then be used if people are planning to write multiple definitions, e.g., "Romanian/relational database terms".

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+3. **Add one new term or definition per Pull Request.** We'd love to receive multiple contributions from you but please keep each separate new term or definition in a distinct branch (see below) and Pull Request. This makes it much easier for the Glosario maintainers/editors to review your contributions. When you open the Pull Request you can refer to the Issue you opened before (see step 2),  e.g. if your Issue was number 49 you should write "Closes \#49." GitHub will then automatically close the Issue if and when your Pull Request is merged.
```
gvwilson, line 15: Again, I suspect this is too heavyweight: my experience is that I usually need to add 2-4 terms at a time because (for example) "anti-join" and "outer join" depend on "join". Creating three PRs would be possible in this case, but one-per-term would _not_ work in the case where definitions refer to each other: if A references B and B references A and they're both new, a single-term PR will fail a glossary check.

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+
+-   The value associated with the `slug` key identifies the entry.
+    -   It must be unique within the glossary.
+    -   It must be in lower case and use only letters, digits, and the underscore
+        (to be compatible with Jekyll's automatic slug creation).
+-   The entry *may* have a `ref` key.
+    If it is present,
+    its value must be a list of identifiers of related terms in this glossary.
+-   Every other top-level key must be a two-letter ISO 639 language code such as `en` or `fr`.
+  (Refer to the "639-1" column of [this table][iso639-table-en].)
```
gvwilson, line 44: indentation

CONTRIBUTING.md
```diff
@@ -0,0 +1,73 @@
...
+    -   Every entry must have at least one such language section.
+-   Within each language section for each term:
```
gvwilson, line 46: Note that the keys like `term` and `acronym` are _not_ to be translated: they must be as-is. (We've had at least one person use the Spanish word for 'term'.)

gvwilson (commented):
My sense is that issue-then-PR-for-every-individual-term is too heavyweight, and that contributions will be easier if people are allowed to submit small clusters of related terms.

CONTRIBUTING.md
tobyhodges: The thinking behind this was to reduce redundancy of effort if many people want to contribute simultaneously (as could be the case during Hacktoberfest): it can be frustrating to spend time working on something only to find someone else was doing exactly the same thing. (This has happened to me before.) Although I hope this project attracts many contributions far into the future, I doubt such measures will be necessary permanently.

tobyhodges:
I'm going to leave the decision on whether to require contributors to 

- open an Issue for each contribution, and 
- create a separate branch/PR for each new term/defintion

up to @gvwilson @zkamvar @ian-flores and @fmichonneau. I'll watch this thread and happily adjust the guide to reflect whatever policy you settle on.

CONTRIBUTING.md
zkamvar: > Again, I suspect this is too heavyweight: my experience is that I usually need to add 2-4 terms at a time because (for example) "anti-join" and "outer join" depend on "join". Creating three PRs would be possible in this case, but one-per-term would _not_ work in the case where definitions refer to each other: if A references B and B references A and they're both new, a single-term PR will fail a glossary check.

I think that's a fair point. Maybe rewording this to say "Add one or a small cluster of related terms per pull request" with an upper limit of, say, ten terms? Since we are using this as a pedagogical tool, I want to reinforce the concept of making smaller pull requests to reduce the burden on both contributors and editors.

fmichonneau (commented):
1. I think that typically contributors to definitions do not spend a lot of time working on them or their associated translation so the likelihood that two people work on the same thing at the same time is pretty low. 
2. while having one entry per pull request makes things easier to review, there are situations where having multiple makes more sense. 

I would suggest that we don't ask for one PR = one issue = one definition, but encourage people who want to work on many terms (e.g., "I want to create definitions for all the terms seen in the tidyverse") to create an issue to signal to others the terms they are working on.

tobyhodges:
In f3fcfdd I've tried to address the points made in discussion above. Can you all check and tell me what you think?

fmichonneau (approved):
Thank you Toby. We can adjust these guidelines depending on the feedback we get but it seems that this is a great starting point.

elletjies:
@all-contributors please add @tobyhodges for doc,question,translation

PR #95, opened 2020-09-27
Adding terms for HTTP

gvwilson:
## Author: 

- @gvwilson 

## Language: 

- English

## Terms defined:

- client
- dom
- dom_selector
- element
- empty_element
- html
- http
- http_header
- http_request
- http_response
- markup_language
- mime_type
- protocol
- query_string
- server
- url

## Editor

@zkamvar 

Depends on #88 and #94.

PR #94, opened 2020-09-27
Terms related to design patterns

gvwilson:
## Author: 

- @gvwilson 

## Language: 

- English

## Terms defined:

- abstract_method
- cache
- caching
- child_class
- design_pattern
- generator_function
- iterator_pattern
- object
- off_by_one_error
- parent_class
- singleton_pattern
- template_method_pattern
- two_hard_problems
- visitor_pattern

## Editor

@zkamvar 

## Depends

Requires #88 for `for_loop`.

PR #96, opened 2020-09-27
More programming terms

gvwilson:
## Author: 

- @gvwilson 

## Language: 

- English

## Terms defined:

- affordance
- asynchronous
- blob
- block_comment
- build_stale
- callback
- console
- cryptographic_hash_function
- dynamic_loading
- dynamic_lookup
- dynamic_scoping
- exception_handler
- globbing
- lexical_scoping
- license
- lifecycle
- line_comment
- lisp
- log
- log_message
- multi_threaded
- non_blocking_execution
- psf
- python
- r_language
- sandbox
- side_effect
- single_threaded
- stack_overflow
- stream
- synchronous

## Editor

@zkamvar 

Needs #88, #94, and #95 for completeness.

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -65,6 +65,13 @@
       if the two are different,
       the test [fails](#fail_test).
 
+- slug: affordance
+  en:
+    term: "affordance"
+    def: >
+      A property of something that suggests how it can be used, such as a handle
+      or button.
+
 - slug: aggregation
   en:
     term: "aggregation"
@@ -195,6 +202,16 @@
     term: "associative array"
     def: >
       See [dictionary](#dictionary).
+
+- slug: asynchronous
+  ref:
+    - synchronous
+  en:
+    term: "asynchronous"
+    def: >
+      Not happening at the same time. In programming, an asynchronous operation
+      is one that runs independently of another, or that starts at one time and
+      ends at another.
 
 - slug: attribute
   en:
@@ -358,6 +375,23 @@
       A unit of information representing representing alternatives, yes/no,
       [true](#true)/[false](#false). In computing a state of either 0 or 1.
 
+- slug: blob
+  en:
+    term: "binary large object"
+    acronym: "BLOB"
+    def: >
+      Data that is stored in a database without being interpreted in any way,
+      such as an audio file. The term is also now used to refer to data
+      transferred over a network as uninterpreted bits.
+
+- slug: block_comment
+  en:
+    term: "block comment"
+    def: >
+      A [comment](#comment) that spans multiple lines. Block comments may be
+      marked with special start and end symbols, like `/*` and `*/` in C and
+      its descendents, or each line may be prefixed with a marker like `#`.
+
 - slug: boilerplate
   en:
     term: "boilerplate"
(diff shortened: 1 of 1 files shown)


glossary.yml
```diff
@@ -358,6 +375,23 @@
...
       [true](#true)/[false](#false). In computing a state of either 0 or 1.
 
+- slug: blob
+  en:
+    term: "binary large object"
+    acronym: "BLOB"
+    def: >
+      Data that is stored in a database without being interpreted in any way,
+      such as an audio file. The term is also now used to refer to data
+      transferred over a network as uninterpreted bits.
```
zkamvar, line 385: Should there be a mention of this usage in git since it's common to see this term in a github URL?

glossary.yml
```diff
@@ -456,6 +490,14 @@
       A specification for a [build manager](#build_manager) that describes how
       some files depend on others and what to do if those files are out of date.
 
+- slug: build_stale
```
zkamvar, line 493: Should the slug be `stale_build` since the term is "stale (in build)"? This would follow the convention of [`child_tree`](https://glosario.carpentries.org/en/#child_tree) referring to a "child (in a tree)".

glossary.yml
```diff
@@ -1091,11 +1161,39 @@
...
+      on the [call stack](#call_stack) at the moment the lookup is done. Almost
+      all programming languages use [lexical_scoping](#lexical_scoping) instead,
+      since it is more predictable.
+
 - slug: emacs
   en:
     term: "Emacs (editor)"
     def: >
-      FIXME
+      A [Lisp](#lisp)-based operating system disguised as a text editor that is
```
zkamvar, line 1195: I'm reading this as a bit of an in-joke among Emacs users that may not necessarily translate for a more general user. Can you rephrase this?

glossary.yml
```diff
@@ -2764,6 +2990,12 @@
       element of the source vector, i.e., `result[index[i]] = source[i]`.  Push
       indexing can easily produce gaps and collisions.
 
+- slug: python
+  en:
+    term: "Python"
+    def: >
+      A popular open-source programming language.
```
zkamvar, line 2997: This definition felt a bit empty, and I think this may be a good addendum (but needs interpreted and compiled to be defined)

```suggestion
      A popular interpreted open-source programming language that relies on white space indentation to define control structure.
```

glossary.yml
```diff
@@ -3952,9 +4246,9 @@
   en:
     term: "XML"
     def: >
-      A set of rules for defining HTML-like tags and using them to format documents
-      (typically data). XML achieved license plate popularity in the early 2000s,
-      but its complexity led many programmers to adopt [JSON](#json) instead.
+      A set of rules for defining HTML-like tags and using them to format
+      documents (typically data). XML was popular in the early 2000s, but its
```
zkamvar, line 4250: This answers https://github.com/carpentries/glosario/pull/95#discussion_r497675293, but will unfortunately clash with #95

zkamvar (dismissed):
Overall, good, but I have some questions around slug syntax and definition expansions.

glossary.yml
gvwilson: Added - thanks.

glossary.yml
gvwilson: All of the other terms related to builds have `build_` as a prefix, so I stuck with that. We should probably discuss/standardize - I could see `child_tree` being `tree_child`.

glossary.yml
gvwilson: Yeah, sorry about that - fixed.

glossary.yml
gvwilson: Added.

glossary.yml
gvwilson: I made the same change to #95.

fmichonneau (approved):

PR #52, opened 2020-08-15
Translate to portuguese: path related terms

villares:
## Author: 

@villares - Alexandre Villares

## Language: 

- Portuguese (from Brazil), used pt for Portuguese. 

## Terms defined:

-  absolute path 
-  relative path
-  current working directory
   (note: the abbreviated "working directory" form seems more common in pt, so that's what I used,
    I can change this if requested.)

## Editor

 @fmichonneau and @beatrizmilz 
  
  I would like to suggest an improvement to the CWD definition in English, mentioning the acronym and that it is often used in the context of CLI (command line interfaces). May I open another pull request for that?

PR #98, opened 2020-10-01
 CONTRIBUTING.md  Add a small phrase about YAML

beatrizmilz:
Hi! @tobyhodges asked me to send this PR. This idea came while doing the translation of the blogpost for Hacktoberfest. 

What do you think @tobyhodges ? How can this phrase be more explainfull?

> Bea:  one thing that I have noticed is that in the yaml file it is easy to make a mistake in the indentation. So is something that we could tell in advance, something like: 'When adding a new term or translation, please beware of the indentation on the YAML file:" (not sure if beware is a good word for that, but couldn't think of a better word).

> Toby: That’s a great point. Please could you make a Pull Request to the Contributing Guide on the Glosario GitHub to add this in?

Thanks

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,8 @@ When you're making your contribution(s) to Glosario, please:
 
 To add a new entry, please [fork][forking-guide] the [main Glosario repository][repo] and, on a new branch, add the term and definition to [`glossary.yml`][glossary]. This glossary file is written in [YAML].
 
+When adding a new term or translation, please notice the indentation on the YAML file.
+
 Here is an example of how your glossary entry should be structured:
 
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ When you're making your contribution(s) to Glosario, please:
 
 To add a new entry, please [fork][forking-guide] the [main Glosario repository][repo] and, on a new branch, add the term and definition to [`glossary.yml`][glossary]. This glossary file is written in [YAML].
 
-When adding a new term or translation, please notice the indentation on the YAML file.
+When adding a new term or translation, please take care with the indentation on the YAML file. Indentation is syntactically significant in YAML.
 
 Here is an example of how your glossary entry should be structured:
 


CONTRIBUTING.md
```diff
@@ -17,6 +17,8 @@ When you're making your contribution(s) to Glosario, please:
 
 To add a new entry, please [fork][forking-guide] the [main Glosario repository][repo] and, on a new branch, add the term and definition to [`glossary.yml`][glossary]. This glossary file is written in [YAML].
 
+When adding a new term or translation, please notice the indentation on the YAML file.
```
tobyhodges, line 20:
```suggestion
When adding a new term or translation, please take care with the indentation on the YAML file. Indentation is syntactically significant in YAML.
```

Perhaps something like this? I wonder if it would be beneficial to point contributors towards a simple online YAML validator like http://www.yamllint.com/ too?

beatrizmilz: Hi @tobyhodges , I think your suggestion  of change is good! About the validator, I think we could add but maybe would need to say that people should use to validate, but don't copy the generated yaml.

Ex: I tried validating a part of the code from #52 (just to test the validator)

the original code:
```
  pt:
    term: "caminho relativo"
    def: >
      Um caminho cujo destino é interpretado de maneira relativa a outro
      local, como o [diretório de trabalho](#current_working_directory). Um caminho
      relativo é o equivalente a indicar um destino com termos como "siga em frente" 
      e "vire a esquerda".
  ref: 
    - absolute_path
```

After I clicked "Go", apeard a message of **Valid YAML!**, but the YAML that I pasted in the validator was changed to this code (with different identation that we need): 

```
--- 
pt: 
  def: "Um caminho cujo destino é interpretado de maneira relativa a outro local, como o [diretório de trabalho](#current_working_directory). Um caminho relativo é o equivalente a indicar um destino com termos como \"siga em frente\"  e \"vire a esquerda\".\n"
  term: "caminho relativo"
ref: 
  - absolute_path
```

So if we add a validator, this should be also something to beware... what do you think?

beatrizmilz: Or maybe we can find a validator that does not change the YAML code

tobyhodges: Yes, if we can't suggest a validator that will preserve our chosen level of indentation I think it's better not to mention one at all.

beatrizmilz: Yes.. If changes the identation and people copy and paste it from the validator, it will create more errors

beatrizmilz: https://codebeautify.org/yaml-validator found this option in google that does not seem to change the code

zkamvar: If it helps, the tests on this repository check if the yaml is valid, so if they submit invalid yaml, an error will be reported with information about how it can be fixed.

beatrizmilz: Thanks @zkamvar ! So there is no need to add a external validator, right?

zkamvar: Bingo.

zkamvar (approved):
LGTM

Issue #109, opened 2020-10-07
2020-10-07: add entry counts for each language on index page


fmichonneau:
add the number of entries for each language on the index page

olexandr-konovalov:
I have noticed that since recently https://glosario.carpentries.org/ now shows the number of untranslated terms for each language. Perhaps it can be made to show both counts?

elletjies:
@olexandr-konovalov thank you for your comment. Please feel free to submit a PR :)

PR #110, opened 2020-10-07
Add number of entries for each language

fmichonneau:
proposed solution for #109 

Still needed:
- [ ] currently "entries" is in English for all languages
- [ ] tweaks to the CSS to differentiate it from the rest of the text would be good.

if anyone wants to contribute to these 2 points, please feel free to do so.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -16,8 +16,9 @@
 
 Assign the PR based on the language to the following person:
 
-| Language | Person | 
-|:--------:|:------:|
-| English  | @zkamvar |
-| Spanish  | @ian-flores |
-| French   | @fmichonneau |
+| Language  | Person        | 
+|:---------:|:-------------:|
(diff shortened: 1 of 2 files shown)


baileythegreen:
I think the translation of 'entries' dovetails nicely with a feature I'm working on, which is part-of-speech information in the appropriate language for each page.

I haven't pushed this yet, but I can add 'entries' to the yaml translate file I have and then push to this branch once I've finished? Or is there a better idea?

I can also do the CSS. Do you want me to just make a decision about how it should look, or did you have an idea?

fmichonneau:
Thanks @baileythegreen!

> I haven't pushed this yet, but I can add 'entries' to the yaml translate file I have and then push to this branch once I've finished? Or is there a better idea?

That sounds like a good approach given our current tooling.

> I can also do the CSS. Do you want me to just make a decision about how it should look, or did you have an idea?

That would be great. I don't have a specific idea. I just thought it would be better if the number of entries didn't stand out too much to keep the focus on the language and the introduction to the project.

PR #113, opened 2020-10-07
Add description of glosario in Afrikaans

jsteyn:
## Author: 

jsteyn

## Language: 

Afrikaans 

## Terms defined:

Definition

## Editor

@jsteyn

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -17,7 +17,15 @@ languages:
     name: Afrikaans
     title: Woordelys
     blurb: >
-      In translation.
+      `glosario` is 'n oop bron woordelys van terminologie wat in datawetenskap 
+    gebruik word en beskikbaar gestel word aanlyn sowel as 'n programmateek in 
+    [R](https://github.com/carpentries/glosario-r/) en [Python]
+    (https://github.com/carpentries/glosario-py/). Deur woordelyssleutels by 'n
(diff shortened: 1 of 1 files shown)


_config.yml
```diff
@@ -17,7 +17,15 @@ languages:
     name: Afrikaans
     title: Woordelys
     blurb: >
-      In translation.
+      `glosario` is 'n oop bron woordelys van terminologie wat in datawetenskap 
+      gebruik word en beskikbaar gestel word aanlyn sowel as 'n programmateek in 
+      [R](https://github.com/carpentries/glosario-r/) en [Python]
+      (https://github.com/carpentries/glosario-py/). Deur woordelyssleutels by 'n
```
zkamvar, line 23: Markdown will not interpret the Python link as a link unless the square brackets and round brackets are on the same line. 

```suggestion
      [R](https://github.com/carpentries/glosario-r/) en [Python](https://github.com/carpentries/glosario-py/). 
      Deur woordelyssleutels by 'n
```

zkamvar (changes requested):

_config.yml
zkamvar: There should be a button above that says "commit suggestion" you can use that to accept the suggestion without changing it on your account if that makes things easier :)

zkamvar (approved):

zkamvar:
Thank you!

PR #100, opened 2020-10-06
Add definition for R_consortium

beacurious:
## Author: 

- @beacurious

## Language: 

- English

## Terms defined:

- R_consortium
- 
- 

## Editor

@zkamvar

PR #114, opened 2020-10-07
Revert 1 master

jsteyn:
## Author: 

jsteyn

## Language: 

Afrikaans 

## Terms defined:

Definition

## Editor

@zkamvar

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -38,7 +38,7 @@ languages:
     blurb: >
       `glosario` es un glosario de fuente abierta de términos utilizados en la
       ciencia de datos que esta disponible en línea y también como libreria en
-      [R](https://github.com/carpentries/glosario-r/) y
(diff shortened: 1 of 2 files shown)


zkamvar:
Hello @jsteyn,

It looks like there may have been a mix-up with branches on your repository and this PR accidentally deletes an item that was added in #100. 

Looking at your fork, it appears that your master branch has some extra commits beyond this repository, so I don't think there's a good way to do this in this particular PR.

My suggestion is the [burn it all down](https://happygitwithr.com/burn.html) approach. In short:

1. delete your fork of this repository from github and your computer
2. re-fork the repository and clone to your computer
3. create a new branch
4. re-add your changes

This is not an uncommon situation to be in (I've done this many times), so you're in good company :)
Do let me know if you would like any help with this.

PR #119, opened 2020-10-07
issue #115: Adding a definition in Arabic to the existing terms

BatoolMM:
## Terms defined:

- compiler
- argument

## Editor

I only added the Arabic translation of two existing terms which are `compiler` and `argument`. As I said previously, it's a bit tricky to write Arabic characters in YAML file since Arabic is written from right to left, which can easily mess up the indentation. If this one worked fine, I'm happy to add a few terms on daily basis.  

Assign the PR based on the language to the following person:

| Language  | Person        | 
|:---------:|:-------------:|
| Afrikaans | @jsteyn       |
| English   | @zkamvar      |
| Spanish   | @ian-flores   |
| French    | @fmichonneau  |

PR #121, opened 2020-10-08
add arabic

fmichonneau:
support for #119

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -26,10 +26,14 @@ languages:
       die inligting kan vind. Outeurs kan ook die programmateekfunksies gebruik om
       konsekwente hiperskakels, vir terme en definisies in enige van verskeie tale, 
       by hul lesse in te sluit.
+  - key: ar
+    name: Arabic
+    title: قائمة المصطلحات
+    blurb: >-
+      In translation.
   - key: en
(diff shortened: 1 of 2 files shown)


fmichonneau:
@gvwilson could you please review and merge this PR and then #119?

_config.yml
```diff
@@ -26,10 +26,14 @@ languages:
       die inligting kan vind. Outeurs kan ook die programmateekfunksies gebruik om
       konsekwente hiperskakels, vir terme en definisies in enige van verskeie tale, 
       by hul lesse in te sluit.
+  - key: ar
+    name: Arabic
+    title: قائمة المصطلحات
+    blurb: >-
```
zkamvar, line 32: Is there a reason there's a `-` after the `>`?

fmichonneau: I'm just copying what @gvwilson does... he taught me everything I know about YAML. 

But I got curious and it led me to this: https://yaml.org/spec/1.2/spec.html#id2794534

>  Strip
>    Stripping is specified by the “-” chomping indicator. In this case, the final line break and any trailing empty lines are excluded from the scalar’s content.

zkamvar (approved):

PR #123, opened 2020-10-08
Defining two more design patterns

gvwilson:
## Author: 

- @gvwilson 

## Language: 

- English

## Terms defined:

- decorator_pattern
- delegate_pattern

## Editor

@zkamvar


zkamvar (approved):
LGTM

TIL about decorators in Python. I've always seen those weird `@` declarations above function calls and knew they were called "decorators", but the docs were always incomprehensible to me before now.

PR #124, opened 2020-10-08
Add two Afrikaans terms and some references

jsteyn:
## Author: 

jsteyn

## Language: 

Afrikaans

## Terms defined:

abstrakte metode
abstrakte sintaksontledingsboom
- 

## Editor

@jsteyn

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -11,7 +11,7 @@
     term: "68-95-99.7 reël"
     def: >
       Beskryf die beginsel dat 68% van alle waardes binne een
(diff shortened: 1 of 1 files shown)


jsteyn (commented):
Removed some spaces that might be causing the problem

glossary.yml
```diff
@@ -1,4 +1,4 @@
-- slug: 68_95_997_rule
+- Slug: 68_95_997_rule
```
zkamvar, line 1:
```suggestion
- slug: 68_95_997_rule
```

This must be a lowercase 's'

fmichonneau (approved):

PR #136, opened 2020-10-17
Removing english words from Arabic translation  

BatoolMM:
## Author: 

- Batool Almarzouq

## Language: 

- Arabic

## Terms defined: None. 

I noticed that adding eng words within any Arabic sentence changes the order of the words when it appears on the site. The order of the words changes in comparison to the one in the glossary.yml. I tried to inspect the site (HTML), found it the same as glossary.yml. So, I tried removing the eng words from the translation. Personally, I think it's better to add some eng terms when explaining or defining the term. For example, when I mention python, I'd write the Arabic translation and between round bracket in English 
` بايثون (python)`

I'll find a better way to correct it later but I removed the eng words for now.

## Editor

| Language   | Person        | 
|:----------:|:-------------:|
| Afrikaans  | @jsteyn       |
| English    | @zkamvar      |
| French     | @fmichonneau  |
| Portuguese | @beatrizmilz  |
| Spanish    | @ian-flores   |


fmichonneau (approved):

fmichonneau:
please let us know if you find a solution

PR #146, opened 2020-10-20
use 'auto' for dir attribute

zkamvar:
We shouldn't have to require editors for rtl languages to add bdi tags in order to override the settings. Luckily, there is the 'auto' specification that detects the direction of the text: https://www.unicode.org/reports/tr9/#Markup_And_Formatting. This also allows the text on the front page to be justified in the correct direction (see screenshot below)

I've set `dir='auto'` in the HTML tags and removed the `<bdi>` tags in the text according with https://www.unicode.org/reports/tr9/#Markup_And_Formatting

@BatoolMM and @fmichonneau, does this look good?

screenshot of Arabic text reading from right to left


BatoolMM (approved):
Awesome! 
I didn't know you could use `auto` to detect the direction of the text. The front page looks way better now. Thank you a lot @zkamvar .

PR #157, opened 2020-10-24
Display acronyms in term headings

timtomch:
This is not an update to the definitions list, but rather a suggestion to slightly alter the display of terms in the glossary.

I think it might be useful to display acronyms in the term headings. For example, someone looking up JSON might want to search for the acronym instead of JavaScript Object Notation.
I suggest adding the acronym in parentheses after the term heading, e.g.

JavaScript Object Notation (JSON)

## Author: 

@timtomch 

## Language: 

- All

## Editor

@zkamvar ?


fmichonneau (approved):

fmichonneau:
Thank you!

PR #160, opened 2020-10-27
Fixing broken entries/cross-references and updating tool.

gvwilson:
1.  Updated `utils/check-glossary.py` to build forward references from entries in all languages.
2.  Fixed broken cross-references:
    -   Some internal links used the term as the target rather than the slug.
    -   Some internal links used `text](slug)` rather than `[text](#slug)`.
3.  Filled in some entries that had a slug but nothing else.

We should encourage contributors or editors to run `make check` before merging changes.

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -515,7 +515,7 @@
   en:
     term: "Bayes' Rule"
     def: >
-      See [Bayes' Theorem](bayes_theorem).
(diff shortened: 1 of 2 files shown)


fmichonneau (approved):

zkamvar:
we may be able to set a github action to perform the check like we do with yaml lint.

PR #161, opened 2020-10-27
run make check instead of just linting yaml in GitHub Action

fmichonneau:
currently, the GitHub action only runs `yamllint`, this PR makes it run the `make check` target which also includes checks on the content and structure of the YAML file


gvwilson (approved):
:+1: from me

PR #168, opened 2020-10-29
Fix the links in the arabic 

BatoolMM:
## Author: 

- Batool Almarzouq

## Language: 

- Arabic

This is to fix the links as @zkamvar has pointed out.


zkamvar (approved):
LGTM and it renders well on my machine. 

Thank you :)

PR #169, opened 2020-10-29
Added Afrikaans terms

elletjies:


glossary.yml
```diff
@@ -1339,7 +1339,13 @@
     def: >
       O quão bem duas variáveis concordam uma com a outra. A correlação é normalmente
       medida pelo cálculo de um [coeficiente de correlação](#correlation_coefficient),
-      e não significa que precise existir causalidade.
+      e não significa que precise existir causalidade.   
+  af:
+    term: "korrelasie"
+    def: >
+      Hoe sterk twee veranderlikes met mekaar ooreenstem. Korrelasie word gewoonlik gemeet aan
+      die berekening van 'n [korrelasiekoëffisiënt](#korrelasiekoëffisiënt), maar
```
zkamvar, line 1347:
```suggestion
      die berekening van 'n [korrelasiekoëffisiënt](#correlation_coefficient), maar
```

glossary.yml
```diff
@@ -1359,6 +1365,15 @@
       conhecer X permite uma previsão perfeita de Y. Se o coeficiente de correlação
       for 0.0, conhecer X não diz nada a respeito de Y, e se for -1.0, então X prevê
       Y, mas mudanças no valor de X gera uma mudança oposta em Y.
+  af:
+    term: "korrelasiekoëffisiënt"
+    def: >
+      'n Maatstaf van hoe sterk [gekorreleer](#korrelasie) twee veranderlikes is. As die
```
zkamvar, line 1371:
```suggestion
      'n Maatstaf van hoe sterk [gekorreleer](#correlation) twee veranderlikes is. As die
```

glossary.yml
```diff
@@ -1359,6 +1365,15 @@
...
+      [korrelasiekoëffisiënt](#korrelasiekoëffisiënt) tussen X en Y gelyk is aan 1.0,
```
zkamvar, line 1372:
```suggestion
      korrelasiekoëffisiënt tussen X en Y gelyk is aan 1.0,
```

elletjies:
@zkamvar thank you for the edits! Much appreciated!

zkamvar:
> @zkamvar thank you for the edits! Much appreciated!

No worries! the errors from yamllint were strange and a bit hard to follow, but it appears that my edits helped. I've been seeing this type of error a lot (attempting to link with the language-specific term, not the slug) and I'm wondering how it can be clear that linking needs to use the slug. 

Additionally, in the future, it would be good to use the issue template to show what words are being added so that auditing PRs is a bit easier (this case was simple, but we've run into issues where _some_ contributors who shall not be named decided to add 50+ terms while listing only 10 :unamused:).

jsteyn (approved):

PR #187, opened 2020-10-31
General editing and corrections

baileythegreen:
#### Note
(I apologise for doing all of this in one pull request, but for general proofreading, the end product comes out much more consistent if it is done all at once. I realised after I finished that the few entries I expanded should be separate. I am happy to create a new branch, copy those changes over, and submit a separate pull request so that they can be reviewed, if that is preferable.)

# Pull Request
Proofreading, additions, and insertion of internal links throughout the file. Where statements were incorrect, misleading, or incomplete, I have tried to correct that. (Again, sorry for not doing this separately.) Also some statements that were opinion- rather than fact-based have been removed, as have phrasings that were intended to be humourous, but which may cause undue confusion for non-native speakers, and colloquial phrases. 

## Author:
- Bailey Harrington, @baileythegreen

## Language: English
(Some small changes made to French or Spanish to remove opinionated statements, but these are very few in number, and no new definitions were added in these languages.)

## Terms defined:

- Nothing new is defined in this pull request.

## Editor


fmichonneau (approved):
wow! Fantastic work @baileythegreen :heart_eyes: I'll let @zkamvar do the final approval and merge but thank you so much for all the improvements your PR brings!

baileythegreen:
Thanks, @fmichonneau! Believe it or not, I actually learned some stuff reading through it, too.

(I'm also really glad someone took the time to go through that giant pull request.)

baileythegreen:
It is maybe worth pointing out that many of the translations appear to be direct translations of the existing English definitions (so errors and the odd opinionated comment are propagated). Ideally where I've made actual corrections, or expanded definitions so that they don't focus on a particular language, the translations could also be updated.

(I know that is a time commitment, and I'd be happy to take a stab at it in the case of French and Spanish entries, with the understanding that a native speaker would look over them after.

fmichonneau:
> It is maybe worth pointing out that many of the translations appear to be direct translations of the existing English definitions (so errors and the odd opinionated comment are propagated). Ideally where I've made actual corrections, or expanded definitions so that they don't focus on a particular language, the translations could also be updated.

Yes, we'd like to build a tool that helps with this issue.

> (I know that is a time commitment, and I'd be happy to take a stab at it in the case of French and Spanish entries, with the understanding that a native speaker would look over them after.

That would be great, and we'd be happy to check these entries.

glossary.yml
```diff
@@ -4356,14 +4423,14 @@
...
-      divided by the desired value. For example, if the actual value is 9 and the
+      divided by the correct value. For example, if the actual value is 9 and the
       correct value is 10, the relative error is 0.1. Relative error is usually more
       useful than [absolute error](#absolute_error).
   pt:
     term: "erro relativo"
     def: >
       O valor absoluto da diferença entre um valor observado e o valor correto,
-      dividido pelo vaor desejado. Por exemplo, se o valor observado é 9 e o correto é
+      dividido pelo vaor correto. Por exemplo, se o valor observado é 9 e o correto é
```
fmichonneau, line 4433: @beatrizmilz could you please check this change?

baileythegreen: @beatrizmilz For context: the change was made to reflect one made in the English definition. The word 'desired' was used, but in a statistical context, there are actual values, and there are correct, or expected, values. Desire has no place here.

glossary.yml
```diff
@@ -2826,7 +2875,7 @@
     term: "k-nearest neighbors"
     def: >
       A [classification](#classification) algorithm that classifies data points based
-      on their similarity to nearby neighbors.
+      on their similarity to their *k* nearest neighbors.
```
zkamvar, line 2878:
```suggestion
      on their similarity to their nearby neighbors
```

glossary.yml
```diff
@@ -2826,7 +2875,7 @@
     term: "k-nearest neighbors"
     def: >
       A [classification](#classification) algorithm that classifies data points based
```
zkamvar, line 2877:
```suggestion
      A [classification](#classification) algorithm that classifies data points into *k* groups based
```

glossary.yml
```diff
@@ -3839,15 +3909,15 @@
...
+      An American comic strip by Charles M. Schulz which has inspired the names of [R](#r_language) versions.
 
 
 - slug: perceptron
   en:
     term: "perceptron"
     def: >
       The simplest kind of [neural network])(#neural_network), which approximates a
-      single neuron with N binary inputs by computing a weighted sum of its inputs and
+      single neuron with $N$ binary inputs by computing a weighted sum of its inputs and
```
zkamvar, line 3920: I don't think we have LaTeX support quite yet.

```suggestion
      single neuron with _N_ binary inputs by computing a weighted sum of its inputs and
```

glossary.yml
```diff
@@ -3941,7 +4014,7 @@
   en:
     term: "prerequisite"
     def: >
-      Something that a [build target](#build_target) depends on.
+      Something that a [build target](#build_target) depends upon.
```
zkamvar, line 4017: Unnecessary

```suggestion
      Something that a [build target](#build_target) depends on.
```

glossary.yml
```diff
@@ -3987,25 +4060,19 @@
...
   en:
     term: "process"
     def: >
       An [operating system](#operating_system)'s representation of a running program.
-      A process typically has some memory, the identify of the user who is running it,
+      A process typically has some memory, the identity of the user who is running it,
       and a set of connections to open files.
 
 
-- slug: product_manager
```
zkamvar, line 4002: I would say it's better not to delete definitions, even if they are poorly defined because it would break backwards compatibility. Instead, add [FIXME] to the definition so that it's easier to identify as problematic.

glossary.yml
```diff
@@ -4028,9 +4096,9 @@
   en:
     term: "prompt"
     def: >
-      The text printed by a [REPL](#repl) or [shell](#shell) that indicates it is
+      The text printed by an [REPL](#repl) or [shell](#shell) that indicates it is
```
zkamvar, line 4099:
```suggestion
      The text printed by a [REPL](#repl) or [shell](#shell) that indicates it is
```

glossary.yml
```diff
@@ -4273,8 +4341,8 @@
   en:
     term: "recycle"
     def: >
-      To re-use values from a shorter vector in order to generate a sequence of the
-      same length as a longer one.
+      To re-use values from a shorter [vector}(#vector) in order to generate a sequence of the
```
zkamvar, line 4344:
```suggestion
      To re-use values from a shorter [vector](#vector) in order to generate a sequence of the
```

glossary.yml
```diff
@@ -4308,13 +4376,13 @@
     term: "regular expression"
     def: >
       A pattern for matching text, written as text itself. Regular expressions are
-      sometimes called "regexp", "regex", or "RE", and are as powerful as they are cryptic.
+      sometimes called "regexp", "regex", or "RE", and are powerful tools for working with text.
```
zkamvar, line 4379: [comment] NICE SAVE!

glossary.yml
```diff
@@ -4566,25 +4639,25 @@
     acronym: "RSE"
     def: >
       Someone whose primary responsibility is to build the specialized software that
-      other researchers depend on.
+      other researchers depend upon.
```
zkamvar, line 4642:
```suggestion
      other researchers depend on.
```

glossary.yml
```diff
@@ -4613,13 +4686,13 @@
   en:
     term: "scalar"
     def: >
-      A single value of a particular type, such as 1 or "a". Scalars do not really
-      exist in R; values that appear to be scalars are actually vectors of unit length.
+      A single value of a particular type, such as 1 or "a". Scalars exist in other languages, but do not really
```
zkamvar, line 4689:
```suggestion
      A single value of a particular type, such as 1 or "a". Scalars exist in several languages, but do not really
```

glossary.yml
```diff
@@ -5137,7 +5227,7 @@
       defines an overall sequence of operations by calling [abstract
       methods](#abstract_method) that [child classes](#child_class) must then
       implement. Each child class then behaves in the same general way, but implements
-      the steps in different ways.
+      the steps in differently.
```
zkamvar, line 5230:
```suggestion
      the steps differently.
```

glossary.yml
```diff
@@ -5180,18 +5270,18 @@
   en:
     term: "tibble"
     def: >
-      A modern replacement for R's data frame, which stores tabular data in columns
+      A modern replacement for [R's](#r_language) data frame, which stores tabular data in columns
```
zkamvar, line 5273: Consistent with earlier examples of this

```suggestion
      A modern replacement for [R](#r_language)'s data frame, which stores tabular data in columns
```

glossary.yml
```diff
@@ -5332,23 +5422,23 @@
...
 
 
 - slug: two_hard_problems
   en:
     term: "two hard problems in computer science"
     def: >
-      Refers to a quote by Phil Karlton: "There are only two hard problems in computer
-      science: cache invalidation and naming things." Many variations add a third
+      Refers to a quote by Phil Karlton–"There are only two hard problems in computer
+      science–cache invalidation and naming things." Many variations add a third
```
zkamvar, line 5441: Needs to be an em-dash, not en-dash

```suggestion
      science—cache invalidation and naming things." Many variations add a third
```

glossary.yml
```diff
@@ -5332,23 +5422,23 @@
...
+      red-green-blue color specification. Tuples are immutable (their values can not be reset.)
 
 
 - slug: two_hard_problems
   en:
     term: "two hard problems in computer science"
     def: >
-      Refers to a quote by Phil Karlton: "There are only two hard problems in computer
-      science: cache invalidation and naming things." Many variations add a third
+      Refers to a quote by Phil Karlton–"There are only two hard problems in computer
```
zkamvar, line 5440:
```suggestion
      Refers to a quote by Phil Karlton—"There are only two hard problems in computer
```

glossary.yml
```diff
@@ -5441,23 +5531,24 @@
...
-    - keyword_arguments
+    - keyword_argument
     - named_argument
   en:
     term: "variable arguments"
     def: >
-      In a function, the ability to take any number of arguments. R uses `...` to
-      capture the "extra" arguments.
+      In a function, the ability to take any number of arguments. [R](#r_language) uses `...` to
+      capture the "extra" arguments. [Python](#python) uses `*args` and `**kwargs` to capture
```
zkamvar, line 5546: Shouldn't `**kwargs` be part of keyword arguments instead of named arguments?

glossary.yml
```diff
@@ -5560,27 +5651,28 @@
...
 
 - slug: vim
   en:
     term: "Vim (editor)"
     def: >
-      The default text editor on Unix. "How do I exit the Vim editor?" is one of the
-      most popular questions on [Stack Overflow](#stack_overflow).
+      The default text editor on Unix. Vim is a very powerful text editor that allows
+      the user to execute shell commands and use [regular expressions](#regular_expression) to
+      alter files programmatically.
```
zkamvar, line 5664: I say this as a user of Vim for the past three years: I think it is definitely worth keeping some sense of how difficult it is for newcomers to use Vim.

zkamvar (changes requested):
Thank you so much for this contribution! Going through the changes, it looks like it was definitely needed. 

In particular, I appreciate the inclusion of more common examples and removing some of the colloquial or snarky language that snuck in there. I've gone through and suggested some changes that fix typos, add context, or revert changes that I felt were not necessary. There are also some questions that I have about a few definitions.

glossary.yml
```diff
@@ -461,11 +473,14 @@
...
     term: "assertion"
     def: >
-      A [Boolean](#boolean) expression that must be true at a certain point in a
+      A [Boolean](#boolean) expression that must be [true](#true) at a certain point in a
       program. Assertions may be built into the language (e.g., [Python](#python)'s
-      `assert` statement) or provided as functions (e.g., R's `stopifnot`). They are
+      `assert` statement) or provided as functions (e.g., [R's](#r_language) `stopifnot`). They are
       often used in testing, but are also put in [production code](#production_code)
-      to check that it is behaving correctly.
+      to check that it is behaving correctly. Assertions should not be used to perform
```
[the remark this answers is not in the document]
baileythegreen: I think R may be an exception here. I don't know, I don't use it. And, to be clear, I am saying assertions should not be used for **data validation** in production code. Using them for testing is fine, but in production code they should only identify things that truly break code. *There are other ways to validate data.*

I feel very strongly about this. [Here is a good summary of why; see Caveat #1: Don't Use Asserts for Data Validation](https://dbader.org/blog/python-assert-tutorial).

Python, and several other languages have an option to optimise assertions away. (I consulted an actual computer science professor on that point when I was correcting this entry, just to make sure this wasn't the minority case.) In Python, assertions are on by default, but turned off under any optimisation condition, but in Java, this is the other way around—assertions must be explicitly enabled (source, Deitel & Deitel 10th edition, Java: How to Program). But, basically, if you rely on data validation to verify the output of a program, and that data validation can be disabled, that seems like really bad practice, particularly if anyone else uses the code and that point isn't very clearly communicated.

There is also a Python lesson that also teaches assertions, and advocates using them for data validation, but that doesn't make it right. I have submitted an [issue](https://github.com/swcarpentry/python-novice-inflammation/issues/874) about it, talked to the maintainers in the issue comments, and am working on a pull request to fix it.

I would be willing to compromise with something citing R as an exception, or saying that 'In many languages, assertions should not be used...,' but I think the point is important to mention.

glossary.yml
baileythegreen: That's what I tried to type, but it wasn't working—my keyboard + vim didn't want to do them. But it looked like the colons were wreaking havoc with the syntax highlighting, and I thought that might pose other problems.

glossary.yml
baileythegreen: Kind of a difficult question. In Python `named arguments` are not a thing, but `keyword arguments` are, and they must be named. I am unsure if this is simply a difference in terminology, or if there are some languages that have both `named arguments` and `keyword arguments`.

I put it here, because this is the entry for `variable (numbers of) arguments`, and `**kwargs` is how a variable number of `keyword arguments` is listed in a Python function definition.

If `**kwargs` needs to be mentioned in `keyword arguments` then, by the same logic, `**args` needs to be mentioned in the entry for `argument`.

glossary.yml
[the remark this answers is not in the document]
zkamvar: After reading the post, I think R is indeed an exception because R does not have the concept of an assert statement in the same sense as other languages since there is no way to ignore these commands with a flag to the interpreter.

R is a bit of an odd language probably because it was built by statisticians. I think it's safe to make a note citing R as an exception to the rule or even saying that R does not have an assert statement.

glossary.yml
zkamvar: Agreed that it is a difficult question. I'm fine to leave this as-is until someone can come up with a clean solution.

glossary.yml
baileythegreen: That is fine. I also use Vim. But what was there before:

```
"How do I exit the Vim editor?" is one of the most popular questions on
[Stack Overflow](#stack_overflow).
```

is unnecessarily negative. I, for one, have never been stuck in Vim, but for AGES, could only exit nano by closing the terminal window.

I don't want the gist of the entry to be like a lot of responses to [this twitter thread](https://twitter.com/cybercdh/status/1318887202680721409) where someone asked about cool Vim shortcuts, and like 2/3 of the replies are just ways to exit Vim (and open nano).

What about:

```
Vim is a very powerful text editor with a steeper learning curve than nano,
but that allows the user to execute shell commands and use
[regular expressions](#regular_expressions) to alter files programmatically.
```

glossary.yml
```diff
@@ -907,8 +922,9 @@
   en:
     term: "catch (an exception)"
     def: >
-      To accept responsibility for handling an error or other unexpected event. R
-      prefers "handling a condition" to "catching an exception".
+      To accept responsibility for handling an error or other unexpected event. [R](#r_language)
+      prefers "handling a condition" to "catching an exception". [Python](#python), on the other hand,
+      encourages raising and catching exceptions, and in some situations, requires it.
```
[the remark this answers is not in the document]
baileythegreen: In truth, "condition handling" is something I have never heard of in programming. But my background is in Python/Java, and I actively avoid R. That sentence was already there, and since I don't know nuances in R, I left it alone and just explained the preferred way in Python.

I did google 'condition handling R' and it mostly sounds the same as what I would call exception handling. I have found reference to both `conditions` and `exceptions` in R, but I don't see how they inherently differ (but they must, right?).

glossary.yml
zkamvar: Oof... that thread is not fun and I agree that the original text was passive-aggressive.

I think your suggested text manages expectations well.

glossary.yml
```diff
@@ -907,8 +922,9 @@
   en:
     term: "catch (an exception)"
     def: >
-      To accept responsibility for handling an error or other unexpected event. R
-      prefers "handling a condition" to "catching an exception".
+      To accept responsibility for handling an error or other unexpected event. [R](#r_language)
+      prefers "handling a condition" to "catching an exception". [Python](#python), on the other hand,
```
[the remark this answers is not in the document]
baileythegreen: Unless I am mistaken, this is going to be flagged by the linter because `handle_condition` is in both the `ref` list, and linked internally. I had to remove a bunch of links I added in for this reason.

(It is possible I accepted some such links already, without realising they were removed from my pull request to satisfy the linter.)

glossary.yml
```diff
@@ -2334,13 +2374,13 @@
   en:
     term: "global environment"
     def: >
-      The [environment](#environment) that holds top-level definitions in R, e.g.,
+      The [environment](#environment) that holds top-level definitions in a programming language, e.g.,
```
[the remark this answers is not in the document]
baileythegreen: Nope. I'd imagine it is this way for any interpreted scripting language.

glossary.yml
```diff
@@ -2442,7 +2483,7 @@
   en:
     term: "group"
     def: >
-      To divide data into subsets according to some criteria while leaving records in
+      To divide data into subsets according to some criterion while leaving records in
```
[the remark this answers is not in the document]
baileythegreen: Of course. I was forgetting that it is possible to group based on multiple fields.

glossary.yml
baileythegreen: I was so disappointed by that thread, because I was so excited to maybe learn something new.

glossary.yml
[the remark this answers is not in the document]
baileythegreen: Yes, I wanted to say 'R is generally a poor choice as a *model* of a programming language, but some R people don't like it when I say things like that.

(i'm glad you clarified there is not an assert statement in R, because I was looking in my books for any reference, and not finding anything.)

glossary.yml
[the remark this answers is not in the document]
zkamvar: I think you are correct! Forgive the intrusion.

glossary.yml
```diff
@@ -2694,13 +2750,6 @@
       A test that checks whether the parts of a system work properly when put together.
```
zkamvar, line 2752:
```suggestion
 - slug: interpreter
   en:
     term: "interpreter"
     def: >
       A program that runs other programs interactively. [FIXME]

```

zkamvar (approved):

glossary.yml
```diff
@@ -2735,13 +2791,11 @@
...
 
-
-- slug: interpeter
-  en:
-    term: "interpeter"
-    def: >
-      A program that runs other programs interactively.
-
+ - slug: interpreter
```
zkamvar, line 2794: Note to past Zhian: YAML is not a markdown list.

```suggestion
- slug: interpreter
```

glossary.yml
```diff
@@ -2735,13 +2791,11 @@
...
-  en:
-    term: "interpeter"
-    def: >
-      A program that runs other programs interactively.
-
+- slug: interpreter
+   en:
+     term: "interpreter"
+     def: >
+       A program that runs other programs interactively. [FIXME]
```
zkamvar, line 2798:
```suggestion
  en:
    term: "interpreter"
    def: >
      A program that runs other programs interactively. [FIXME]
```

remove whitespace

zkamvar:
Thank you again @baileythegreen for the submission!

baileythegreen:
@zkamvar Before you consider this 'finished', I think you missed the comments I left last night on the suggestions I didn't close. You've just pushed a version with two entries for 'interpreter', an inaccurate definition of 'k-nearest neighbours', and two other things that I care less about in the REPL and scalar entries.

zkamvar:
> @zkamvar Before you consider this 'finished', I think you missed the comments I left last night on the suggestions I didn't close. You've just pulled a version with two entries for 'interpreter', an inaccurate definition of 'k-nearest neighbours', and two other things that I care less about in the REPL and scalar entries.

Sorry about that. I didn't see any comments on those items and I assumed that they were okay to change. 

For reference, here are the individual comments, which I have also marked as unresolved:

interpreter: https://github.com/carpentries/glosario/pull/187#discussion_r516116980
K-nearest neighbors: https://github.com/carpentries/glosario/pull/187#discussion_r516128422, https://github.com/carpentries/glosario/pull/187#discussion_r516128606
REPL: https://github.com/carpentries/glosario/pull/187#discussion_r516136447
scalar: https://github.com/carpentries/glosario/pull/187#discussion_r516142073

For example, I do not see a comment on K-nearest neighbors beyond my initial suggestion. I get a feeling that GitHub's interface may have let us down here :/

image

(and yes, I now realize the error I introduced :facepalm:. I swear I actually can brain most days...)

Do you want to create a new PR or just comment on the issue comments?

baileythegreen:
@zkamvar I can do a separate pull request with those four changes (interpreter, k-nearest neighbours, REPL, and scalar). That will probably be simpler. This one should have been more than one to begin with, and is now quite long and convoluted. I may do this tomorrow as it is quite late here.

ps. I am super jealous that you live in Portland. I love it there.

elletjies:
@all-contributors please add @baileythegreen for doc,review,question,translation

PR #198, opened 2020-11-04
Translate argument, ascii, assertion to afr. Add fields to empty slug…

jsteyn:
Translate argument, ascii, assertion to afr. Add fields to empty slug

## Author: 

- jsteyn

## Language: 

- Arikaans

## Terms defined:

- argument
- ascii
- assertion

## Editor

| Afrikaans  | @jsteyn, @elletjies |
| French     | @fmichonneau        |

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -3,10 +3,9 @@
     term: "68-95-99.7 Regel"
     def: >
       Drückt die Tatsache aus, dass 68% der Werte innerhalb einer
-      [Standardabweichung](#standard_deviation) vom Mittelwert liegen,
-      95% innerhalb von zwei und 99,7% innerhalb von drei. Umgekehrt
-      liegen etwa 0,3% der Werte in den meisten Fällen mehr als drei
-      Standardabweichungen über oder unterhalb des Mittelwerts.
+      [Standardabweichung](#standard_deviation) vom Mittelwert liegen, 95% innerhalb
+      von zwei und 99,7% innerhalb von drei. Umgekehrt liegen etwa 0,3% der Werte in
+      den meisten Fällen mehr als drei Standardabweichungen über oder unterhalb des Mittelwerts.
   en:
     term: "68-95-99.7 rule"
     def: >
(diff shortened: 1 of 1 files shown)


jsteyn:
My editor does not like empty slugs, so I had to add the basic fields to those, eg:
- slug: someword
  en: ""
  def: >

The editor also reformats all lines to not exceed 80 characters.

I hope these changes are okay.

glossary.yml
```diff
@@ -40,14 +39,10 @@
...
     def: >
-        يُعبِر عن حقيقة أن ٦٨٪ من القيم تقع ضمن وحدة واحدة من
-        [الانحراف المعياري](#standard_deviation)
-        من
-        [الوسط](#mean)
-        ، بينما ٩٥٪ تقع ضمن وحدتين،
-        و٩٩,٧٪ تقع ضمن ثلاث وحدات
-        بالمقابل ، حوالي ٠,٣ ٪  من القيم تقع  في أكثر من ثلاثة وحدات
-        من الانحرافات المعيارية أعلى أو أقل من المتوسط في معظم الحالات.
+      يُعبِر عن حقيقة أن ٦٨٪ من القيم تقع ضمن وحدة واحدة من [الانحراف
```
fmichonneau, line 42: I think your text editor is changing the order of the words and breaking the links here.

glossary.yml
```diff
@@ -78,16 +73,15 @@
   ar:
     term: "البرمجيات المهجورة أو المتروكة"
     def: >
-       هي البرامج التى لايتوفر لها دعم من أي برمجيات أُخرى
-       ولم تعد على قيد الصيانة.
+      هي البرامج التى لايتوفر لها دعم من أي برمجيات أُخرى ولم تعد على قيد الصيانة.
```
fmichonneau, line 76: same here the order of the words is being changed.

glossary.yml
```diff
@@ -336,8 +325,8 @@
     term: "desenvolvimento ágil"
     def: >
       Um método de desenvolvimento de software que enfatiza vários passos pequenos e
-      feedback contínuo ao invés de planejamento de longo prazo. [Programação exploratória](#exploratory_programming)
-      costuma ser ágil.
+      feedback contínuo ao invés de planejamento de longo prazo. [Programação
```
fmichonneau, line 328: it would also be best if the link elements were not broken up

fmichonneau (changes requested):
We need to come up with a technical solution to ensure consistent formatting of the pull requests to avoid issues with different text editor configurations. In the case of your pull request, it seems that the re-wrapping of the entries is breaking the links for the Arabic entries. Could you please update your pull request to only include the entries you have added? Thanks!

zkamvar:
I agree with @fmichonneau's comments, but given that this PR is > 1K lines of change, I would suggest instead to close this, and open a new fresh branch.

jsteyn:
I'll close this issue and create a new branch making the changes as suggested.

PR #199, opened 2020-11-04
Removing unneeded ref to mean in definition of median.

gvwilson:
The definitions of `median` include an explicit reference to `mean` so it isn't needed under the `ref` sub-key.

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -3396,7 +3396,6 @@
 
 - slug: median
   ref:
-    - mean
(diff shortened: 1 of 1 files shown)


zkamvar (approved):

zkamvar:
I think it is a bit confusing when we should place a term in ref and when we should make an explicit reference in the definition. At the moment, it seems to be dealer's choice.

PR #200, opened 2020-11-04
Removing redundant definition of 'interpreter'

gvwilson:
The term `interpreter` was defined twice - this PR removes one and adds a few words to the other.


zkamvar (approved):
The redundant addition was 100% my fault. Thank you for fixing it.

PR #201, opened 2020-11-04
Adding 'de' to list of accepted languages in utilities

gvwilson:
`make check` was reporting `de` as an unknown language. (Every time a new language is added, `utils/check-glossary.py` must be updated accordingly.)


zkamvar (approved):
[infomercial voice]
There's got to be a better way!
[/infomercial voice]

PR #204, opened 2020-11-04
Closes #203

### LINKED ISSUE #203: 2020-11-4: Fix issues subsequent to #187
For @baileythegreen.

Fix entries for:

- interpreter (duplicate entry)
- k-nearest neighbours (incorrect definition)
- REPL (grammar issue)
- scalar (wording change)

baileythegreen:
## Author: 

- Bailey Harrington, @baileythegreen 

## Language: 

- English

## Terms defined:

- Alterations made to:
    - interpreter (removed duplicate entry)
    - k-nearest neighbours (corrected definition)
    - REPL (fixed a grammar issue)
    - scalar (changed wording)

as outlined in #203.

## Editor

For @zkamvar, who this has already been discussed with.

| Afrikaans  | @jsteyn, @elletjies |
| Arabic     | @BatoolMM           |
| French     | @fmichonneau        |
| Portuguese | @beatrizmilz        |
| Spanish    | @ian-flores         |

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -3391,7 +3391,6 @@
 
 - slug: median
   ref:
-    - mean
     - mode
   en:
     term: "median"
--- a/glossary.yml
+++ b/glossary.yml
@@ -2791,11 +2791,6 @@
     def: >
       A test that checks whether the parts of a system work properly when put together.
 
-- slug: interpreter
-  en:
-    term: "interpreter"
-    def: >
-      A program that runs other programs interactively. [FIXME]
 
 - slug: interpreted_language
   en:
@@ -2923,8 +2918,8 @@
   en:
     term: "k-nearest neighbors"
     def: >
-      A [classification](#classification) algorithm that classifies data points into *k* groups based
-      on their similarity to their nearby neighbors
+      A [classification](#classification) algorithm that classifies data points based
+      on their similarity to their *k* nearest neighbours. 
 
 
 - slug: kebab_case
@@ -4175,7 +4170,7 @@
   en:
     term: "prompt"
     def: >
-      The text printed by a [REPL](#repl) or [shell](#shell) that indicates it is
+      The text printed by an [REPL](#repl) or [shell](#shell) that indicates it is
       ready to accept another command. The default prompt in the Unix shell is usually
       `$`, while in [Python](#python) it is `>>>`, and in [R](#r_language) it is `>`.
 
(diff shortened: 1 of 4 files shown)


glossary.yml
```diff
@@ -4169,7 +4171,7 @@
   en:
     term: "prompt"
     def: >
-      The text printed by a [REPL](#repl) or [shell](#shell) that indicates it is
+      The text printed by an [REPL](#repl) or [shell](#shell) that indicates it is
```
zkamvar, line 4174: Mostly curious: why use 'an' when the beginning of the acronym does not have a vowel sound?

```suggestion
      The text printed by a [REPL](#repl) or [shell](#shell) that indicates it is
```

glossary.yml
```diff
@@ -2791,6 +2791,8 @@
     def: >
       A test that checks whether the parts of a system work properly when put together.
 
+
+
```
zkamvar, line 2795:
```suggestion
```

zkamvar: Full disclosure, this was my bad

glossary.yml
gvwilson: I think it depends on whether you're pronouncing the acronym as a word or spelling it out: "a sequel database" versus "an ess cue ell database".

zkamvar: One of these days, my brain will work properly.... but probably not for a while.

zkamvar (approved):

zkamvar:
Thank you again @baileythegreen! It looks good from here. Sorry for the holdup.

glossary.yml
baileythegreen: @zkamvar, @gvwilson  Here is the comment I originally left regarding [a REPL/an REPL] and why the second is correct. It is a difference between the word (acronym, in this case) starting with a **vowel sound** versus starting with a **vowel**. (I think all words that start with a vowel also start with a vowel sound, because I can't think of any counterexamples; however, not all words that start with a vowel sound start with a vowel. I am uncertain as to the size of that list, though.)

#### Original comment
I understand why you would think this, but because the actual pronunciation of the letter 'R' starts with a vowel sound, it needs the 'n'.

(If you say R-E-P-L, the first syllable is the same as the first syllable of argument, and you say an argument.

zkamvar: @baileythegreen Thank you for the clarification. I did understand the difference between using an with vowel sounds, but never heard REPL spelled out before. 

This also brings up a good point that potential clashes between American and British rules (e.g. an herb garden vs. a herb garden) and it means that we either need to choose one flavor/flavour of English or forever be adding or subtracting single letter changes. 

There's a reason I love and hate the English language all the same. :weary:

baileythegreen: @zkamvar I say this as an American who has lived in Britain for the last five years, and long time basher of English as a language: English is ridiculous. It makes no sense. But i think British English makes even less sense because they often randomly just don't pronounce letters.

I think it is possible that the 'herb' issue might be singular. I don't know of another word where the determination of whether it starts with a vowel sound versus not is dependent upon the speaker's accent. Luckily, 'herb' is unlikely to appear in this glossary.

PR #208, opened 2020-11-07
Fixing four typos

gvwilson:
- "optimization" instead of "optimisation" (American spelling)
- "mathematician" instead of "mathemetician"
- "asynchronously" instead of "ascynrhonously"
- "supposed" instead of "suposed"

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -518,7 +518,7 @@
       often used in testing, but are also put in [production code](#production_code)
       to check that it is behaving correctly. In many languages, assertions should not be
       used to perform data-validation as they may be silently dropped by compilers and interpreters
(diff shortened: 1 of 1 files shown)


zkamvar (approved):
LGTM. We should probably specify what grammar and spelling conventions we want to abide by.

baileythegreen:
I think I introduced at least that one instance of British spelling without thinking about it. Probably worth considering what sort of spellchecker might be used with English (pursuant to #209) as if a method is used that only does American spellings, it would be silly to set British as the standard.

fmichonneau:
yes, British spelling is preferred in The Carpentries.

PR #221, opened 2020-11-20
define: function

TomKellyGenetics:
## Author: 

- Tom Kelly

## Language: 

- English

- Japanese

## Terms defined:

- function

- 関数

## Editor

| Japanese | @masamiy |


glossary.yml
```diff
@@ -1749,6 +1749,22 @@
...
+  en:
+    term: "function"
+    def: >
+      A code block which gathers a sequence of operations into a whole, preserving it for ongoing use
+      by a defining tasts that take a set of inputs and generate expected outputs. They enable
+      repeating these defined tasks with one command.
+  ja:
+    term: "関数"
+    def: >
+      連続した操作を一つに纏めようにコードブロックとして後から使うときのために保存しておきます。
```
naoe-tatara, line 1765: "纏めように" does not sound very natural. I would suggest the following: "別の処理の中で使用する、あるいは繰り返し使用するために、連続した操作をひとかたまりのコードブロックとして名前を付けて定義したもの。"

glossary.yml
```diff
@@ -1749,6 +1749,22 @@
...
+      関数は繰り繰返し演算を実行できるのように入力と返した出力を定義します。
```
naoe-tatara, line 1766: Also, "実行できるのように" does not need "の".

naoe-tatara (changes requested):
Hi, I would suggest some changes in Japanese translation. See the comments I made at specific lines. Please let me know if there is anything unclear.

TomKellyGenetics:
Thanks for the feedback! I think this an improvement, I've updated the PR with the suggested changes.

glossary.yml
masamiy: Agreed with naoe-tatara:
```suggestion
      関数は繰り繰返し演算を実行できるように入力と返した出力を定義します。
```

masamiy (changes requested):
Thank you for the contribution! 
Will you please have a look at suggested changes?

baileythegreen:
I'm going to venture a guess that it should say 'tasks'.

glossary.yml
```diff
@@ -1749,6 +1749,22 @@
...
 
+- slug: function
+  ref:
+    - functions
+    - command
+  en:
+    term: "function"
+    def: >
+      A code block which gathers a sequence of operations into a whole, preserving it for ongoing use
+      by a defining tasts that take a set of inputs and generate expected outputs. They enable
```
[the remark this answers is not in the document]
TomKellyGenetics: Well spotted. It's "tasks" typed badly 👍

glossary.yml
TomKellyGenetics: I agree to this change. I think I got confused with the Kanji for "iterative" (operation).

```diff
-関数は繰り繰返し演算を実行できるように入力と返した出力を定義します。
+関数は繰り返し演算を実行できるように入力と返した出力を定義します。
```

I'm not sure about the other point. I haven't mentioned 引数 for "argument" (also called option, flag, or parameter), although I think this would a correct translation in Japanese. I actually left this out of the English definition as well for simplicity and consistency. 

I'm trying to be consistent with the translated R lessons here:

https://swcarpentry-ja.github.io/r-novice-gapminder/10-functions/index.html
https://swcarpentry-ja.github.io/r-novice-gapminder/ja/_episodes/10-functions/index.html

Here we have the following definition where "出力" refers to the output. So my intention here is "返した出力" as "the output(s) that is/are returned".
 * a defined set of inputs and expected outputs
* 入力と出力の組み合わせを定義

Unfortunately we haven't translated the relevant section for Shell or this would be helpful.

https://swcarpentry-ja.github.io/shell-novice/01-intro/index.html

TomKellyGenetics: I've added a "optional arguments" to both definitions to clarify this.

masamiy: Thanks for the clarification. What about just '出力', or '返された出力' or 'リターンされた出力'? It could be just me feeling 返した出力 does not sound right, so, I need a third opinion - What do you think, @naoe-tatara ?

TomKellyGenetics: Ah I see. Passive form makes sense to me.  Of course, as a non-native speaker it is good for me to know if it's unnatural Japanese. Wikipedia for [サブルーチン](https://ja.wikipedia.org/wiki/%E3%82%B5%E3%83%96%E3%83%AB%E3%83%BC%E3%83%81%E3%83%B3#%E5%91%BC%E7%A7%B0) / [Function](https://en.wikipedia.org/wiki/Function_(computer_science)) uses 返り値 for "returned value".

glossary.yml
```diff
@@ -1749,6 +1749,23 @@
...
 
+- slug: function
+  ref:
+    - functions
+    - command
+  en:
+    term: "function"
+    def: >
+      A code block which gathers a sequence of operations into a whole, preserving it for ongoing use
+      by a defining tasks that take a set of inputs and optional arguments to return expected outputs.
```
baileythegreen, line 1760:
```suggestion
      by defining a set of tasks that takes zero or more required and optional arguments as inputs and return expected outputs (return values), if any.
```

Elsewhere in this glossary we (read: I) have defined all of the input values passed to a function as arguments, which can be either required or optional, and with the understanding that a function can take required arguments, optional arguments, a mixture of the two, or no arguments at all. Consequently, this wording in English feels inconsistent.

Return values are also optional in some languages (Python, for example, has an optional `return` statement, and will implicitly return `None` if no `return` statement or return value are given).

@zkamvar  Thoughts on this altered English wording? (Note, I've also fixed a typo.) Also, the tense of 'to take' here depends on whether a group of objects is treated as a single thing, or several, and the US/K environments I live in are conflicted here.

glossary.yml
```diff
@@ -1749,6 +1749,23 @@
...
+      They enable repeating these defined tasks with one command.
```
baileythegreen, line 1761:
```suggestion
      Functions enable repeating these defined tasks with one command, known as a function call.
```

baileythegreen (changes requested):
@TomKellyGenetics I've added some comments on the English version that provide a better fit with other related definitions.

glossary.yml
TomKellyGenetics: Thanks that's helpful to know (as of course I'm new to the project). If it's possible to be consistent with other terms that would be ideal!

This is a more difficult one to define that I'd initially expected. Different programming languages have slightly different terminology for these concepts. @masamiy has actually raised similar issues with the Japanese version with the added confusion of English loanwords or Chinese characters. In hindsight, a bilingual contribution was ambitious. Thanks for checking it.

glossary.yml
```diff
@@ -1749,6 +1749,24 @@
...
+- slug: function
+  ref:
+    - functions
+    - command
+  en:
+    term: "function"
+    def: >
+      A code block which gathers a sequence of operations into a whole, preserving it for ongoing use
+      by defining a set of tasks that takes zero or more required and optional arguments as inputs
+      and return expected outputs (return values), if any. Functions enable repeating these defined
```
baileythegreen, line 1761:
```suggestion
      and returns expected outputs (return values), if any. Functions enable repeating these defined
```

I missed an s! Darn collective noun verb conjugations....

zkamvar (approved):

glossary.yml
TomKellyGenetics: Updated for consistency with English version:

* returns expected outputs (return values), if any.
* もしあれば返された出力(返り値)を定義します。

masamiy: Thanks @TomKellyGenetics !

TomKellyGenetics:
@masamiy Thanks for your feedback. I think it's improved a lot.

I wasn't sure whether to use Instead of 任意の or 任意な or 任意的な.

Could you explain your other suggestion? It's not clear what the benefit of this to me.

baileythegreen (approved):

baileythegreen:
The English looks good now!

TomKellyGenetics:
> 関数は定義されたタスクを、関数呼び出しと呼ばれる１つのコマンドで繰り返し使用することを可能にします。

@masamiy is this to match the [changes](
https://github.com/carpentries/glosario/pull/221/commits/fa52e7b25e4363589fa25871149e12361fcf3bf7) to the English version? With addition of this clause?

"Functions enable repeating these defined tasks with one command, **known as a function call**."

```diff
-by a defining tasks that take a set of inputs and optional arguments to return expected outputs.
-They enable repeating these defined tasks with one command.
+by defining a set of tasks that takes zero or more required and optional arguments as inputs and return expected outputs (return values), if any.
+Functions enable repeating these defined tasks with one command, known as a function call.
```

Do you want this entire sentence added?

```diff
関数は繰り返し演算を実行できるように、任意の引数を使って、入力ともしあれば返された出力(返り値)を定義します。
+関数は定義されたタスクを、関数呼び出しと呼ばれる１つのコマンドで繰り返し使用することを可能にします。
```
I have another idea.



```diff
+関数呼び出しと呼ばれるときに、
関数は繰り返し演算を実行できるように、
任意の引数を使って、
入力ともしあれば返された出力(返り値)を定義します。
```

naoe-tatara:
Hi, I am sorry for not being to able to contribute so far, but I like the (2) best; personally this feels most natural.

TomKellyGenetics:
No problem, it's good to get a second opinion on this. I've updated the PR for this one. 

ありがとうございます！

masamiy:
Hi @TomKellyGenetics and @naoe-tatara, Thank you for interpreting what I was trying to add and making a beautiful paragraph about function. Yes, I meant to match the changes to the English version with addition of the clause. I like the {2) best.

baileythegreen:
@masamiy @naoe-tatara 

Are any of your requested changes still pending? If not, can you change your review status? There's another PR whose ability to pass the linter checks depends on this PR's acceptance because it references `function` in an internal link.

TomKellyGenetics:
@fmichonneau @tobyhodges @baileythegreen I think it's safe to go ahead and merge this. Both Japanese reviewers seem to be happy with the new version.

baileythegreen:
I am not sure about the others, but I can't merge this until the review requesting changes is resolved.

masamiy (approved):

PR #222, opened 2020-11-20
Add quotes around terms and two line-ends after an entry

jsteyn:
General edit to add quotes around terms and two line-ends after an entry


fmichonneau:
Thanks @jsteyn. What is the motivation for adding quotes around the acronyms? If it's stylistic, I think that would be best handled in the template itself as different languages use different quote formats.

fmichonneau (approved):

PR #237, opened 2020-12-02
Closes #234: Add DRY Principle.

### LINKED ISSUE #234: 2020-12-01: Add "DRY principle"
We are linking to glossario for the [Jekyll pages lesson](https://carpentries-incubator.github.io/building-websites-with-jekyll-and-github-or-gitlab/reference.html) and thought it might be good to add a definition for the `DRY principle`.  I'm not confident in my definition and would appreciate someone else drafting it. Happy to help or review though.

[Draft PR adding glassario links to glossary page](https://github.com/carpentries-incubator/building-websites-with-jekyll-and-github-or-gitlab/pull/165)

baileythegreen:
Add entry for DRY Principle.

## Author: 

- Bailey Harrington

## Language: 

- English

## Terms defined:

- DRY principle

## Editor

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -2029,7 +2029,7 @@
       A vote against something.
 
 
-- slug: dry_principle
+- slug: dry
   en:
     term: "DRY principle"
     def: >
--- a/glossary.yml
+++ b/glossary.yml
@@ -2031,9 +2031,9 @@
 
 - slug: dry_principle
   en:
-    term: "DRY Principle"
+    term: "DRY principle"
     def: >
-      The Don't Repeat Yourself (DRY) Principle states that — Every piece of knowledge
+      The Don't Repeat Yourself (DRY) principle states that — Every piece of knowledge
       must have a single, unambiguous, authoritative representation within a system.
       The term comes from The Pragmatic Programmer, by Andrew Hunt and David Thomas.
       Programs that follow the DRY Principle avoid duplication of definitions and logic,
--- a/glossary.yml
+++ b/glossary.yml
@@ -2029,6 +2029,18 @@
       A vote against something.
 
 
+- slug: dry_principle
+  en:
+    term: "DRY Principle"
+    def: >
+      The Don't Repeat Yourself (DRY) Principle states that — Every piece of knowledge
+      must have a single, unambiguous, authoritative representation within a system.
+      The term comes from The Pragmatic Programmer, by Andrew Hunt and David Thomas.
+      Programs that follow the DRY Principle avoid duplication of definitions and logic,
+      so that a change in their behaviour only requires each modification to be made in
+      one part of the code. The goal is to create code that is easier to maintain.
+
+
 - slug: dynamic_loading
   en:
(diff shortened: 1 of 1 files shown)


baileythegreen:
@zkamvar

In addition to reviewing the definition, maybe look at the structure of the entry. Since this is a term that **contains** but is not **only** an acronym, I am not positive of the correct way to handle it. However, I believe that using the `acronym:` key in the `glossary.yml` file would format it like the API entry, which doesn't quite work.

Screen Shot 2020-12-02 at 14 50 45

zkamvar:
> @zkamvar
> 
> In addition to reviewing the definition, maybe look at the structure of the entry. Since this is a term that **contains** but is not **only** an acronym, I am not positive of the correct way to handle it. However, I believe that using the `acronym:` key in the `glossary.yml` file would format it like the API entry, which doesn't quite work.
> 
> Screen Shot 2020-12-02 at 14 50 45

My gut feeling is to drop the "_principle" part in the slug and have the term be "Don't Repeat Yourself" with the acronym "DRY" (a la the API slug). This does clash with `68_95_997_rule` in that the role of the definition is no longer included, but there is precedent for DRY being used as a stand-alone descriptor (e.g. [make sure you have DRY code](https://www.drycode.io/)).

baileythegreen:
Another thing: this will have some side effects whatever we choose: for the part-of-speech stuff, it changes whether this would be considered a noun or an adjective.

zkamvar:
> Another thing: this will have some side effects whatever we choose: for the part-of-speech stuff, it changes whether this would be considered a noun or an adjective.

:thinking: I think DRY exists as both a noun and an adjective (e.g. the noun form allows the title [DRY is for losers](https://www.theserverside.com/discussions/thread/39358.html#203288)).

What if we do something like this:

```yaml
- slug: dry
  en:
    term: "Don't Repeat Yourself Principle"
    acronym: DRY
```

Whatever we decide, I feel like the slug should be `dry` since it is much easier to spell and type than `dry_principle`. What say you @baileythegreen?

baileythegreen:
I think this might be slightly preferable:

```
- slug: dry
  en:
    term: "Don't Repeat Yourself Principle"
    acronym: "DRY"
```

Quotes because all of the other acronyms have them.

This should render to:

`Don't Repeat Yourself Principle (DRY)`, correct? That seems fine.

glossary.yml
```diff
@@ -2029,6 +2029,18 @@
       A vote against something.
 
 
+- slug: dry_principle
```
zkamvar, line 2032:
```suggestion
- slug: dry
```

glossary.yml
```diff
@@ -2029,6 +2029,18 @@
...
+  en:
+    term: "DRY principle"
```
zkamvar, line 2034:
```suggestion
    term: "Don't Repeat Yourself principle"
    acronym: "DRY"
```

zkamvar (changes requested):

zkamvar (approved):

PR #244, opened 2020-12-08
hard code repo URL instead of using the GitHub API

fmichonneau:
Relying on the GitHub API made the site at https://glosario.carpentries.org broken.
Using the API also makes it more difficult for contributors as additional steps are needed to make the site work without error message. Given that the API is only used here to return the URL of the repository, it's easier to hard code the URL and to directly reference it in the code.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,5 +1,5 @@
 title: "Glossary"
-repository: <USERNAME>/<PROJECT>
+repository_url: "https://github.com/carpentries/glosario"
 defaults:
   - scope:
       path: ""
--- a/_includes/title.html
+++ b/_includes/title.html
@@ -14,7 +14,7 @@
 <div class="row" dir="auto">
   <div class="left">
     <a href="{{'/' | relative_url}}"><img class="logo" src="{{'/static/parrot.svg' | relative_url}}" alt="Parrot logo"/></a>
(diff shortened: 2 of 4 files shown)


baileythegreen:
To clarify, because this came up in an issue recently, https://glosario.carpentries.org just displays the raw html.

I think the proposed changes have the potential to make it simpler for people to deploy locally, as I for one, always needed to explicitly specify a username/repo in the config file for it to deploy. I'll test this before approving it, though.

fmichonneau:
Thanks @baileythegreen! Yes `glosario.carpentries.org` is currently broken. To be able to use this URL, I set up building the site on Netlify where the magic `repository: <USERNAME>/<REPO>` doesn't work as it's GitHub-specific. I hadn't noticed that it was broken before because there was no alert on deployment failure. I fixed that.

zkamvar (approved):
LGTM. I like the aspect of making this more portable and always wondered what the heck was going on with that repository config variable.

fmichonneau:
@baileythegreen let me know when you have gotten a chance to test the proposed change.

baileythegreen (approved):
Was able to serve the site locally without having to modify anything; hooray! Should make it easier for people new to Jekyll, et al. to contribute.

fmichonneau:
thanks!

PR #255, opened 2020-12-15
Add 'clustering' in Japanese and change style

masamiy:
1. Added a Japanese translation in 'clustering'.
2. Changed 'masu' to '... no koto' for 'git'.

## Author: 

- @masamiy 

## Language: 

- Japanese

## Terms defined:

- Clustering
- Git entry minor change
- 

## Editor

| Afrikaans  | @jsteyn, @elletjies          |
| Arabic     | @BatoolMM                    |
| Japanese   | @tomkellygenetics, @masamiy  |
| Portuguese | @beatrizmilz                 |
| Spanish    | @ian-flores                  |

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -1348,6 +1348,10 @@
     def: >
       O processo de dividir os dados em grupos quando os grupos em si não são conhecidos
       antecipadamente.
+  ja:
+    term: "クラスタリング"
+    def: >
+      データをどのように分類するかは不明のまま、グループに分割するプロセスのこと。
 
 
 - slug: code_coverage
@@ -2552,7 +2556,7 @@
   ja:
     term: "Git(ギット)"
     def: >
-       プロジェクトなどの変更履歴を記録・追跡するための分散型バージョン管理システムです。
+       プロジェクトなどの変更履歴を記録・追跡するための分散型バージョン管理システムのこと。
   pt:
     term: "Git"
     def: >


masamiy:
I have assigned @naoe-tatara as a reviewer (sorry, but you are the only one!). I would greatly appreciate it if you have a look at this whenever you have time (no rush :)). Thank you so very much and Merry Christmas :)

glossary.yml
```diff
@@ -1348,6 +1348,10 @@
     def: >
       O processo de dividir os dados em grupos quando os grupos em si não são conhecidos
       antecipadamente.
+  ja:
+    term: "クラスタリング"
+    def: >
+      データをどのように分類するかは不明のまま、グループに分割するプロセスのこと。
```
naoe-tatara, line 1354: This seems a perfect translation of the English definition here in glossary.yml. However, Given the first paragraph of INTRODUCTION of ['A new approach to clustering' (1969, Ruspini)](https://doi.org/10.1016/S0019-9958(69)90591-9) on p. 23,  I feel the English definition here misses information about how the grouping is done as a result (but not "how" as means, if this makes sense). So IMO, the English definition should be something like "The process of dividing data into groups when the groups themselves are not known in advance, but so that as a result data points within a group are more similar to each other than than data points in the other groups in certain perspective."

I am not an expert of this field, so I maybe misunderstanding. I felt like the current definition feels like "grouping data in whatever way, could be random". So if this is the case, I think again the translation is perfect.

glossary.yml
```diff
@@ -2552,7 +2556,7 @@
   ja:
     term: "Git(ギット)"
     def: >
-       プロジェクトなどの変更履歴を記録・追跡するための分散型バージョン管理システムです。
+       プロジェクトなどの変更履歴を記録・追跡するための分散型バージョン管理システムのこと。
```
naoe-tatara, line 2559: Awesome! Thanks @masamiy :)

naoe-tatara (commented):
I left 2 comments here. I would approve the one in line 2559 (deleted line number 2555), but want some clarification (not on the translation but) on the English definition which the Japanese translation is based on).

TomKellyGenetics (approved):

naoe-tatara:
@masamiy , do you have any further comments on the translation&definition of 'clustering'?

glossary.yml
TomKellyGenetics: Perhaps I can help here? Clustering is used a lot in my field.

You're right that the result of clustering is to group similar data points together but the "how" complicates things. There are many ways to perform clustering (hierarchical, graph-based, k-means, etc) so it's difficult to generalise and I think not necessarily helpful to novice users to go into detail. I agree that the English definition is very simple but I think it is okay.

masamiy:
@naoe-tatara No - At least Japanese version represents what English says and if it should be more meaningful, a suggestion should be made to the English definition first. Thanks!

naoe-tatara (approved):
Thanks @masamiy and @TomKellyGenetics , then I think this translation should be approved as it is. Sorry for "getting lost" in the GitHub conversation. I noticed the email when Tom replied to my comment, but when I came to this PR from Masami's comment, I could not find Tom's comment.

TomKellyGenetics:
Thanks @naoe-tatara, glad I could help. 😄 

I think GitHub does confusing things with replies to threads. See this https://github.com/carpentries/glosario/pull/255#discussion_r619017921.

elletjies:
@all-contributors please add @masamiy for translation

PR #257, opened 2020-12-15
Add definition for "preamble"

sackmanna:
added a definition for the term "preamble"

## Author: 

- 

## Language: 

- 

## Terms defined:

- 
- 
- 

## Editor

| Afrikaans  | @jsteyn, @elletjies          |
| Arabic     | @BatoolMM                    |
| Japanese   | @tomkellygenetics, @masamiy  |
| Portuguese | @beatrizmilz                 |
| Spanish    | @ian-flores                  |

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -4426,6 +4426,15 @@
     def: >
       A naming style that separates the parts of a name with underscores, as in `first_second_third`.
 
+- slug: preamble
+  ref:
+    - LaTeX
+  en:
+    term: "preamble"
+    def: >
+      A series of commands prior to the \begin{document} command that define the type of document based on attributes and parameters. 
+      This is also the section of the document where packages are added using the command \usepackage{} to enable additional functionalities. 
+
 
 - slug: prerequisite
   ref:
--- a/glossary.yml
+++ b/glossary.yml
@@ -4432,8 +4432,11 @@
   en:
     term: "preamble"
     def: >
-      A series of commands prior to the \begin{document} command that define the type of document based on attributes and parameters. 
-      This is also the section of the document where packages are added using the command \usepackage{} to enable additional functionalities. 
+      A series of commands, either placed in the main document, or kept in a separate document,
+      that are included prior to the \begin{document} command. The preamble defines the type of
+      the document, along with other formatting attributes and parameters. This is also the section
+      of the document where packages are added using the command \usepackage{} to enable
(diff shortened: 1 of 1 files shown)


glossary.yml
```diff
@@ -4426,6 +4426,15 @@
...
       A naming style that separates the parts of a name with underscores, as in `first_second_third`.
 
+- slug: preamble
+  ref:
+    - LaTeX
+  en:
+    term: "preamble"
+    def: >
+      A series of commands prior to the \begin{document} command that define the type of document based on attributes and parameters. 
+      This is also the section of the document where packages are added using the command \usepackage{} to enable additional functionalities.
```
baileythegreen, line 4436:
```suggestion
      A series of commands, either placed in the main document, or kept in a separate document,
      that are included prior to the \begin{document} command. The preamble defines the type of
      the document, along with other formatting attributes and parameters. This is also the section
      of the document where packages are added using the command \usepackage{} to enable
      additional functionalities, and where custom commands can be defined. 
```

baileythegreen (changes requested):
I've expanded the definition because it seemed limited to how a small project would be laid out. I use LaTeX for some very large projects, and I never include the preamble in the main document itself. In my current one, I actually have a directory of files for specifying things like new commands, colours, the title page, and then the preamble.tex file is just a bunch of \input{} statements.

I am checking to see which of the other editors use LaTeX to request a review.

baileythegreen:
Note: In searching to see if the term `preamble` is used in languages other than LaTeX, I am seeing many references to `boilerplate code`. This is essentially the same concept, but I want to **strongly suggest** that we avoid using this term in the definition as it is jargon-y and may not be well-understood by non-native speakers. I have no idea how well it might translate.

zkamvar:
> I am checking to see which of the other editors use LaTeX to request a review.

@baileythegreen Your changes look good to me. I have also used preambles similarly.

ian-flores:
@baileythegreen I don't want to merge before getting your okay because you were running point on this PR. With the merging of your suggestion can we know merge this PR into main?

baileythegreen (approved):

baileythegreen:
@ian-flores Yes, I have just been trying to not be really heavy-handed about my suggestions and automatically commit them.

This was more than a reformatting, by the way. There were pretty significant changes to the definition, but Zhian already signed off on that.

elletjies:
@all-contributors please add @sackmanna  for translation

PR #258, opened 2020-12-16
Update Pull Request template for clarifications.

TomKellyGenetics:
### Suggested changes

Adds the following line to the Pull Request template to avoid notifying all editors.

```diff
+Please remove the lines not relevant to the language submitted.
```

### Proposal

Add @naoe-tatara as Japanese editor if she agrees to it.

### Question for admins

We've discussed style for Japanese contributions and decided to use 体言止め(taigen-dome) which more concise and is commonly dictionaries. We'll submit PRs to update the existing Japanese entries (e.g., PR #255). Is there a style guide to notify new contributors of this or do we bring it up when reviewing each `lang-ja` PR?

### Assign admins and notify Japanese contributors

Assign this PR based on the language to the following person:

| Language   | Person                                   |
| English    | @baileythegreen, @zkamvar   |
| French     | @fmichonneau  |
| Japanese | @masamiy, @naoe-tatara |

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -22,7 +22,7 @@ Assign the PR based on the language to the following person:
 | Arabic     | @BatoolMM                    |
 | English    | @baileythegreen, @zkamvar    |
 | French     | @fmichonneau                 |
-| Japanese   | @tomkellygenetics, @masamiy  |
+| Japanese   | @masamiy, @naoe-tatara       |
 | Portuguese | @beatrizmilz                 |
 | Spanish    | @ian-flores                  |
 
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -25,3 +25,5 @@ Assign the PR based on the language to the following person:
 | Japanese   | @tomkellygenetics, @masamiy  |
 | Portuguese | @beatrizmilz                 |
 | Spanish    | @ian-flores                  |
+
+Please remove the lines not relevant to the language submitted.


baileythegreen:
The addition looks good to me, particularly as I almost submitted the exact same thing the other day....

As for a style guide, it's a good idea, but I don't know of one. If there is one, it's been hidden very well, I think. The Contributing Guidelines focus more on YAML syntax/checking the build of the site than anything language-specific.

Given the number of conversations the Japanese editors have had recently on style, a guide would seem to be a useful thing; there may be other languages that could benefit from it, for instance, languages with gendered nouns, where we might set guidelines on whether specific conventions should be used in that language's definitions to make them gender-neutral (at least with respect to references to people).

We may also want to avoid use of some [computer terms](https://t.co/GQbKUFDVCx?amp=1) to be more inclusive.

naoe-tatara:
Hi, Thanks @TomKellyGenetics , I agree to be added as Japanese editor 👍

TomKellyGenetics:
Thanks @naoe-tatara, I've updated the editors. I'm happy to continue contributing and reviewing but I've removed myself. This is separate to this issue raised above. I'm a non-native Japanese speaker and I think 2 native speaking editors if enough.

@baileythegreen we have some style guides and documentation on @swcarpentry-ja, we could add a glossario page there for Japanese if we're the only ones that need language specific guidance. If possible, I think directing people there (when we have the documentation ready) or having contributors guidelines for each language somewhere for the glosario project.

>We may also want to avoid use of some computer terms to be more inclusive.
I understand the concerns raised here but they're not so relevant to Japanese specifically, almost all words are gender neutral for example.

baileythegreen:
@TomKellyGenetics 

I don't think Japanese is the only language that would benefit from a style guide, so I would be in favour of including them here, and would be happy to build it into the website, too.

My point here:

> We may also want to avoid use of some computer terms to be more inclusive.

was not specific to Japanese, just something I thought of as I typed the answer. I think it's something we should consider generally, though obviously the specifics of how or whether to implement any decision, will be language-dependent and ultimately up to the editors to enact/enforce.

fmichonneau:
Thanks for this!

I invited @naoe-tatara to join the glosario-editor team. 

I think a style guide would be wonderful! I'll open an issue to keep track of it.

fmichonneau (approved):

PR #262, opened 2020-12-21
Add 'data structures'

mdbjax:
Added data_structures

## Author: 

- @mdbjax

## Language: 

- English

## Terms defined:

- data structure

## Editor

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -1813,6 +1813,13 @@
     def: >
       Alguém que usa habilidades de programação para resolver problemas estatísticos.
 
+- slug: data_structure
+  en:
+    term: "data structure"
+    def: >
+      A format for the organization, management, and efficient access of data. Consists of 
+      a set of data values, the relationship between those data values, and a function to read or write
+      to the data that is stored.
 
 - slug: data_wrangling
   en:
--- a/glossary.yml
+++ b/glossary.yml
@@ -1817,9 +1817,10 @@
   en:
     term: "data structure"
     def: >
-      A format for the organization, management, and efficient access of data. Consists of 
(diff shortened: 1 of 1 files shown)


glossary.yml
```diff
@@ -1813,6 +1813,13 @@
     def: >
       Alguém que usa habilidades de programação para resolver problemas estatísticos.
 
+- slug: data_structure
+  en:
+    term: "data structure"
+    def: >
+      A format for the organization, management, and efficient access of data. Consists of 
+      a set of data values, the relationship between those data values, and a function to read or write
+      to the data that is stored.
```
baileythegreen, line 1822:
```suggestion
      A format for the organisation, management, and efficient access of data. Typically it will
      characterise a set of data values and their representation (or [encoding](#character_encoding)),
      the relationships between values, and ways to access or manipulate those data, such as
      reading, altering, or writing.
```

Thanks for this! This is a particularly broad thing to try to define, but an important one. After consultation on your proposed definition with a CS professor, I'm requesting some changes. They keep the sense of your definition, but make it less concrete, so it more accurately describes a greater proportion of all possible data structures.

I've also changed it to British spellings (preferred in this project) and added an internal link.

baileythegreen (changes requested):
Some changes requested to make the definition broader.

baileythegreen:
I'm going to let @zkamvar review this, since I've done some more heavy-handed editing here.

baileythegreen (approved):

zkamvar (approved):
I think this is spot on. I was tempted to add examples, but I realized that they could be added in `ref` once we have those in place.

PR #265, opened 2021-01-04
Update PULL_REQUEST_TEMPLATE.md

baileythegreen:
Despite @TomKellyGenetics's recent changes to the template, it is still not being used as we intend and we're all still getting spammed with notifications.

This way, the handles are all hidden in an html comment and the correct ones must be actively assigned. I've seen this done in lots of other repos that use PR templates.

I also added some instructions for naming PRs that deal with adding terms to hopefully reduce the number of `Update glossary.yml` ones we get.

## Author: 

- @baileythegreen 

## Language: 

- html

## Terms defined:

- none

## Editor


fmichonneau:
Thanks, Bailey! I agree that the current system doesn't work now that we have many languages and editors. I don't think the GitHub Assignment can be done by contributors but only by maintainers/editors. I think it'd make more sense to remove this section from the template and instead create teams for each language which could then be assigned during triaging by other maintainers. What do you think?

baileythegreen:
@fmichonneau  I think not having the assignment in the template is reasonable. My main question about the plan you outline would be: Do we have a master list of editors somewhere else? Do all of the maintainers know who all of the editors are for each language? I am not a maintainer, but without that list in the template, I don't know the language turf of the different editors. And, while this could be put on the website in the Editors section of a contributors page, that isn't super-helpful when you're looking at a PR inside the browser interface.

Do you know a way to resolve that that doesn't need to handles to be in the template? Or, will this not be an issue for the maintainers? (Keeping in mind that the list of languages/editors is only going to grow.)

baileythegreen:
To clarify: I was intending for contributors to copy and paste the handles of the appropriate editors for their PRs, not use the assignment mechanism GitHub provides, because, of course, they can not.

fmichonneau:
I was thinking that when people agree to become editors, we add them to language specific teams that children of the more general Glosario editor teams (e.g. https://github.com/orgs/carpentries/teams/glosario-editors and https://github.com/orgs/carpentries/teams/glosario-editors-en), then we can use these teams as assignees for issues or PR.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -14,8 +25,9 @@
 
 ## Editor
 
-Assign the PR based on the language to the following person:
+- 
 
+<!--Assign the PR based on the language to the correct editor(s):
```
baileythegreen, line 30:
```suggestion
<!--Assign the PR based on the language by tagging the correct editor(s) from this table:
```

baileythegreen:
That probably would be fine. I don't have any experience making or managing GitHub (sub)teams, so I'm not in a position to give anything other than general feedback, but I think having them actually set up in GitHub renders my previous points moot.

baileythegreen:
@fmichonneau Where are we at with this? I saw that some sub-teams have been created for English, French, and Spanish; does that mean that we can remove the tagging table from the PR template? Also, does this mean we can close Issue #266?

fmichonneau:
I created the teams, I think we can now remove the table in the template here.

baileythegreen:
@fmichonneau Thanks. So, do I need to do anything else, or are we just waiting for a positive review from a not-me (and maybe not-you?) person?

fmichonneau (approved):

fmichonneau:
We should be good now!

PR #273, opened 2021-01-27
Fix rendering of Bayes' Theorem equation

zkamvar:
The ever frustrating kramdown parser is to tables what Microsoft Excel is to dates: it thinks anything that has two or more `|` characters is a table, which creates problems for math, which is documented: https://kramdown.gettalong.org/syntax.html#math-blocks

This adds math blocks and replaces `|` with `\vert` to avoid overzealous tabulation.

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -790,14 +790,14 @@
     term: "Bayes' Theorem"
     def: >
       An equation for calculating the probability that something is [true](#true) if something
-      related to it is true. If P(X) is the probability that X is true and P(X|Y) is
-      the probability that X is true given Y is true, then P(X|Y) = P(Y|X) * P(X) / P(Y).
+      related to it is true. If P(X) is the probability that X is true and \$$P(X\vertY)$$ is
+      the probability that X is true given Y is true, then \$$P(X\vertY) = P(Y\vertX) * P(X) / P(Y)$$.
   es:
     term: "Teorema de Bayes"
     def: >
       Una ecuación para calcular la probabilidad de que algo sea [verdadero](#true) si algo    
-      relacionado con ello es verdadero. Si P(X) es la probabilidad de que X is verdadero y P(X|Y) es    
-      la probabilidad de que X es verdadero dado que Y sea verdadero, entonces P(X|Y) = P(Y|X) * P(X) / P(Y).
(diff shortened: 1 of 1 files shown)


zkamvar:
Okay, that didn't work at all :weary:

baileythegreen:
@zkamvar 

> The ever frustrating kramdown parser is to tables what Microsoft Excel is to dates.

It may not have worked, but this is the best commit message I have seen. It's like if a high school student was given an overly-technical TAKS writing prompt. (I don't know if you were in school at the right time for this reference, substitute SAT in, if not.)

'Write about a time when you felt a program acted overzealously, and any negative impact that may have had.'

baileythegreen:
Also, can we not just make the definitions double-quoted strings? (I know single-quoted will cause problems with apostrophes.)

zkamvar:
> Also, can we not just make the definitions double-quoted strings? (I know single-quoted will cause problems with apostrophes.)

This still doesn't work because of how zealously kramdown loves tables. You know what DOES work? putting equations in `code`, It's not great for accessibility though because it will be something like "code P right parenthesis A vertical bar B left parenthesis". But then again, that IS an improvement over the current display, which is accessible to absolutely no one.

baileythegreen:
The answer seems clear: LaTeX!

But really though, what if we rendered the equation elsewhere, properly, in LaTeX, saved the image with that terrifying parrot logo, then devised a way to have the html/Liquid code fetch the image to the correct place upon building the page?

baileythegreen:
Actually, we could just add another heading called `img:` to listings. That would work (so long as we added code to act upon it, too).

baileythegreen:
Also, I literally just made the connection that `kram` in `kramdown` is just `mark` backwards. And it felt important to record this. Here. For posterity.

zkamvar:
> The answer seems clear: LaTeX!
> 
> But really though, what if we rendered the equation elsewhere, properly, in LaTeX, saved the image with that terrifying parrot logo, then devised a way to have the html/Liquid code fetch the image to the correct place upon building the page?

The LaTeX solution was what I was going for in this PR (based on the documentation), but something got borked and I think it may be line breaks, but I have no clue.

zkamvar:
> Also, I literally just made the connection that `kram` in `kramdown` is just `mark` backwards. And it felt important to record this. Here. For posterity.

It's fitting because kramdown's reasoning for things is kinda backwards.

PR #275, opened 2021-02-02
Add 'configuration file' definition.

AiliSarre:
The term "configuration file" was added in English to glossary.yml

## Author: 

- Aili Sarre

## Language: 

- English

## Terms defined:

- configuration file
- 
- 

## Editor

| English    | @baileythegreen |

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -1563,6 +1563,13 @@
       that simulates such a device.
 
 
+- slug: configuration file
+  en:
+    term: "configuration file"
+    def: >
+      A file that specify the parameters and initial settings of software programs. Config files are often used for information subject to changes, such as environment-specific settings.
+      
+      
 - slug: constant
 
 
--- a/glossary.yml
+++ b/glossary.yml
@@ -1567,7 +1567,9 @@
   en:
     term: "configuration file"
     def: >
-      A file that specify the parameters and initial settings of software programs. Config files are often used for information subject to changes, such as environment-specific settings.
+      A file that specifies the parameters and initial settings of a software program.
+      Configuration, or config, files are often used for information subject to changes,
+      such as environment-specific settings.
       
       
 - slug: constant


glossary.yml
```diff
@@ -1563,6 +1563,13 @@
       that simulates such a device.
 
 
+- slug: configuration file
+  en:
+    term: "configuration file"
+    def: >
+      A file that specify the parameters and initial settings of software programs. Config files are often used for information subject to changes, such as environment-specific settings.
```
baileythegreen, line 1570:
```suggestion
      A file that specifies the parameters and initial settings of a software program. Configuration,
      or config, files are often used for information that is subject to change, such as
      environment-specific settings.
```

baileythegreen (changes requested):
I'm requesting some grammatical changes, and splitting the text over multiple lines to keep it from being too wide.

baileythegreen:
@AiliSarre

Thanks. A few more guidelines for future additions:

*  rename the PR based on the term(s) being added, and
*  split the definition over multiple lines so it can be read without horizontal scrolling!

I've requested changes, but I'm leaving them for you to commit, in case you want to discuss them.

glossary.yml
AiliSarre: Thank you @baileythegreen. I have made the suggested changes. However, I was uncertain whether, and why, you also suggest I should do a line change after Configuration, ...?

AiliSarre:
@baileythegreen I am sorry about forgetting the line shift, I remembered just after I had sent the pull-request.

glossary.yml
baileythegreen: The text after my suggestions is only split over 3 lines. `Configuration,`  appears on the first line, but in the suggestion this is wrapped, so might look like it is on a different line. An actual new line will be numbered on the far left, and then have the `+` symbol to indicate a change has been made.

baileythegreen: It doesn't look like you have committed the changes, actually. You can do so by clicking 'Commit suggestion' below it.
Screen Shot 2021-02-03 at 11 41 45

AiliSarre: Thanks for clarifying. I did not commit the changes yet, I wanted to wait till I had heard from you. Thanks for pointing out the "commit suggestion" option, it seems there is a time limit for using that? It does not matter, it is fine that I get to practice this a few times.

baileythegreen: I don't think there is a time limit. That would be weird. Is that option not available to you?

As far as a new PR, no. Once a PR is submitted, you can still make commits on top of it by pushing to the same branch: in this case, add_config_file. This is a sensible thing to do, so long as the new commits are related to the original intent of the PR. (In this case they are.)

AiliSarre: I think perhaps the problem was that I had already committed the changes (commit d56e9fd9dc7aaa428f94083e4ee437bb8e8011e9) before I noticed the "commit suggestion" button. When I hover the mouse there it says "outdated suggestions cannot be applied". 

I really hope everything is in order now. Thank you for your patience with me.

baileythegreen: Oh, I see; yes, that is indeed what happened. For some reason it wasn't clear on my end yesterday. Anyway, in future, if here are specific suggestions, as in this case, you can use that button to commit them, which might be easier.

But if there are larger changes, or just vague comments about changes requested, then the way you did it might be necessary/preferable.

baileythegreen (approved):

PR #290, opened 2021-03-01
added pip and pypi (English)

callumrollo:
## Author: 

- Callum Rollo

## Language: 

- English

## Terms defined:

- PyPI
- pip

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -4483,13 +4483,14 @@
 
 - slug: pip
   ref:
+    - pypi
     - python
   en:
     term: "Pip Install Packages"
     acronym: "PIP"
     def: >
-      The standard [package manager](#package_manager) for [Python](#python). pip Enables the download and installation
-      of Python packages not included in the standard library.
+      The standard [package manager](#package_manager) for [Python](#python). pip enables
+      the download and installation of Python packages not included in the standard library.
 
 
 - slug: pipe_operator
@@ -4754,12 +4755,14 @@
 - slug: pypi
   ref:
     - python
+    - pip
   en:
     term: "Python Package Index"
     acronym: "PyPI"
     def: >
-      The official third-party software repository for [Python](#python). Anyone can upload a [package](#package)
-      to PyPI. PyPI packages may install via executed scripts or pre-compiled system-specific wheels.
+      The official third-party software repository for [Python](#python). Anyone can upload
+      a [package](#package) to PyPI. PyPI packages may install via executed scripts or
+      pre-compiled, system-specific wheels.
 
 
 - slug: python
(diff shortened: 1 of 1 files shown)


glossary.yml
```diff
@@ -4481,6 +4481,17 @@
...
 
+- slug: pip
+  ref:
+    - python
+  en:
+    term: "Pip Install Packages"
+    acronym: "PIP"
+    def: >
+      The standard [package manager](#package_manager) for [Python](#python). pip Enables the download and installation
+      of Python packages not included in the standard library.
```
baileythegreen, line 4492:
```suggestion
      The standard [package manager](#package_manager) for [Python](#python). pip enables
      the download and installation of Python packages not included in the standard library.
```


We try to keep lines from getting too wide—no set character limit, but use some longer entries in the file to see where to put line breaks; but don't split words.

baileythegreen: Also, `enables` didn't need to be capitalised.

glossary.yml
```diff
@@ -4740,6 +4751,17 @@
...
 
+- slug: pypi
+  ref:
+    - python
+  en:
+    term: "Python Package Index"
+    acronym: "PyPI"
+    def: >
+      The official third-party software repository for [Python](#python). Anyone can upload a [package](#package)
+      to PyPI. PyPI packages may install via executed scripts or pre-compiled system-specific wheels.
```
baileythegreen, line 4762:
```suggestion
      The official third-party software repository for [Python](#python). Anyone can upload
      a [package](#package) to PyPI. PyPI packages may install via executed scripts or
      pre-compiled, system-specific wheels.
```

Line breaks again.

glossary.yml
```diff
@@ -4740,6 +4751,17 @@
       easily produce gaps and collisions.
 
 
+- slug: pypi
+  ref:
+    - python
```
baileythegreen, line 4756:
```suggestion
    - python
    - pip
```

glossary.yml
```diff
@@ -4481,6 +4481,17 @@
       Phony targets are often used to store commonly-used commands in a [Makefile](#makefile).
 
 
+- slug: pip
+  ref:
+    - python
```
baileythegreen, line 4486:
```suggestion
    - pypi
    - python
```

baileythegreen (changes requested):
I've suggested some additional crosslinks between the two entries (I *think* these should compile okay, since both are being added simultaneously), and changes to line breaks.

callumrollo (commented):
Incorporating changes from @baileythegreen. Thanks for the improvements! Hopefully the simultaneous cross links will work

baileythegreen (approved):

elletjies:
@all-contributors please add @callumrollo for translation

PR #292, opened 2021-03-08
Fixes to glossary after proof-reading by Taylor & Francis.

gvwilson:
These minor fixes are based on feedback from a professional proof-reader at CRC Press/Taylor & Francis for the *Research Software Engineering with Python* book. None of them change the meaning of definitions, so no changes should be required to languages other than English.

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -5856,7 +5856,7 @@
     term: "tag (in version control)"
     def: >
       A readable label attached to a specific [commit](#commit) so that it can easily
-      be referred to later.
+      be referenced later.
 
 
 - slug: tdd
--- a/glossary.yml
+++ b/glossary.yml
@@ -3212,7 +3212,7 @@
     term: "label (an issue)"
     def: >
       A short textual tag associated with an [issue](#issue) to categorize it. Common
-      labels include `bug` and `feature request`.
+      labels include [`bug`](#bug) and `feature request`.
 
 
 - slug: issue_tracking_system
--- a/glossary.yml
+++ b/glossary.yml
@@ -2086,7 +2086,7 @@
     acronym: "DOI"
     def: >
       A unique persistent identifier for a book, paper, report, dataset, software release, or
-      other digital artifact.
+      other digital artefact.
   ja:
     term: "デジタルオブジェクト識別子"
     acronym: "DOI"
--- a/glossary.yml
+++ b/glossary.yml
@@ -622,10 +622,10 @@
       `assert` statement) or provided as functions (e.g., [R](#r_language)'s `stopifnot`). They are
       often used in testing, but are also put in [production code](#production_code)
       to check that it is behaving correctly. In many languages, assertions should not be
-      used to perform data-validation as they may be silently dropped by compilers and interpreters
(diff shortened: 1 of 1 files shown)


glossary.yml
```diff
@@ -3212,14 +3212,14 @@
     term: "label (an issue)"
     def: >
       A short textual tag associated with an [issue](#issue) to categorize it. Common
-      labels include [`bug`](#bug) and `feature request`.
+      labels include `bug` and `feature request`.
```
baileythegreen, line 3215: Why is the internal link here being removed?

glossary.yml
```diff
@@ -2086,7 +2086,7 @@
     acronym: "DOI"
     def: >
       A unique persistent identifier for a book, paper, report, dataset, software release, or
-      other digital artefact.
+      other digital artifact.
```
baileythegreen, line 2089: @fmichonneau 

You told me British spellings are preferred. Can you confirm this, in which case this change (and at least one other here, to `stylise`) should not be made.

glossary.yml
```diff
@@ -3360,7 +3360,7 @@
     term: "LaTeX"
     def: >
       A typesetting system for document preparation that uses a specialized [markup language](#markup_language) to define a document structure
-      (e.g. headings), stylise text, insert mathematical equations, and manage citations and cross-references. LaTeX is widely
+      (e.g., headings), stylize text, insert mathematical equations, and manage citations and cross-references. LaTeX is widely
```
baileythegreen, line 3363: This is a British vs American spelling; I believe British is preferred, so this should not be changed.

baileythegreen (commented):
I would personally not make most of these changes, but I accept that I use punctuation differently than others, and that the suggestions are fine.

I have added, I think three, comments on changes that conflict with what I have been told about preferences for British vs American spellings, and one change for which I don't understand the need.

glossary.yml
fmichonneau: yes within The Carpentries we favor British spelling

fmichonneau:
@baileythegreen you are the editor for this resource so I let you decide which edits you want to accept/reject.

glossary.yml
baileythegreen:
```suggestion
      other digital artefact.
```

glossary.yml
baileythegreen:
```suggestion
      labels include [`bug`](#bug) and `feature request`.
```

glossary.yml
baileythegreen:
```suggestion
      (e.g., headings), stylise text, insert mathematical equations, and manage citations and cross-references. LaTeX is widely
```

glossary.yml
```diff
@@ -5856,7 +5856,7 @@
     term: "tag (in version control)"
     def: >
       A readable label attached to a specific [commit](#commit) so that it can easily
-      be referenced to later.
+      be referred to later.
```
baileythegreen, line 5859:
```suggestion
      be referenced later.
```

baileythegreen (approved):

PR #296, opened 2021-03-31
Add definition for 'keyword' (English)

leedotson:

## Author: 

- @leedotson

## Language: 

- English

## Terms defined:

- keyword
- 
-

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -3341,9 +3341,10 @@
   en:
     term: "keyword"
     def: >
-      A word which is reserved by a program because it has a special meaning and cannot
-      be used as variable names, function names, or other identifiers. Keywords can
-      be commands or parameters. Also known as reserved names or reserved words.
+      A word which has a special meaning in a programming language and therefore cannot be
+      used as a variable name, a [function](#function) name, or other identifier. Keywords can
+      be [commands](#command) or [parameters](#parameter). Also known as reserved names
+      or reserved words.
       
 
 - slug: keyword_argument


glossary.yml
```diff
@@ -3337,6 +3337,15 @@
       such as a [dictionary](#dictionary).
 
 
+- slug: keyword
+  en:
+    term: "keyword"
+    def: >
+      A word which is reserved by a program because it has a special meaning and cannot
+      be used as variable names, function names, or other identifiers. Keywords can
+      be commands or parameters. Also known as reserved names or reserved words.
```
baileythegreen, line 3346:
```suggestion
      A word which is reserved by a program because it has a special meaning and cannot
      be used as a variable name, a [function](#function) name, or other identifier. Keywords can
      be [commands](#command) or [parameters](#parameter). Also known as reserved names
      or reserved words.
```

baileythegreen (changes requested):
See inline comment. I think the definition needs some changes before it is added.

baileythegreen:
In case it isn't clear, I both suggested specific changes, and described others that need to be made.

leedotson:
@baileythegreen Thanks for the feedback. I was just typing that question and figured it out as you posted.

elletjies (approved):

zkamvar:
Thank you for your contribution, @leedotson. Since this PR is over a year old, I am going to close this. Feel free to reopen with the requested changes. 

I agree with @baileythegreen that the scope of this definition is too broad for the term "keyword." Specifically, a keyword does not necessarily need to be a reserved word (as Bailey described with python's named parameters (aka keyword arguments)).

PR #298, opened 2021-04-07
Adds a definition of the Help system (English/Japanese)

TomKellyGenetics:
Defines the "[Help (Command)](https://en.wikipedia.org/wiki/Help_(command))" in both English and Japanese as ヘルプ(機能).

## Author: 

- @TomKellyGenetics 

## Language: 

- English
- Japanese

## Terms defined:

- Help(Command)
- ヘルプ(機能)


glossary.yml
```diff
@@ -2855,6 +2855,22 @@
       not their data types or units).
 
 
+- slug: help_system
+  ref:
+    - documentation
+    - help_command
+    - help_function
```
baileythegreen, line 2862:
```suggestion
```

The `ref:` portion of these entries should point towards other existing entries, not simply provide synonyms or related terms because they are used to supply internal links at the bottom of the entry. We currently don't have any of these as entries.

glossary.yml
```diff
@@ -2855,6 +2855,22 @@
...
+  en:
+    term: "help (command)"
+    def: >
+      Opens a help page for the command. The system provides detailed information
+      on how the command works and how it is used.
```
baileythegreen, line 2867: I'd like to see this a bit more general; in programming language interpreters help is usually a command, but for things that run via the command line, it is an option handled by the parser. It also seems a bit weird to have the slug be `help_system`, but the term be `help (command)`, particularly when `help_command` is then listed as one of the references.

baileythegreen (changes requested):
Some comments inline.

glossary.yml
TomKellyGenetics: Thanks for pointing that out. I just noticed that when running into failing checks for PR #303. I've removed these for now.

glossary.yml
TomKellyGenetics: I agree that a broader definition would be helpful. I wrote this in the context of the R lessons and as also described for [Help (command)](https://en.wikipedia.org/wiki/Help_(command)) here.

Rather than changing this, I propose to change the slug to `help_command` and leave room for a wider "help (system)" definition.

glossary.yml
```diff
@@ -2855,6 +2855,18 @@
...
+- slug: help_command
+  en:
+    term: "help (command)"
+    def: >
+      Opens a help page for the command. The system provides detailed information
+      on how the command works and how it is used.
+  ja:
+    term: "ヘルプ(機能)"
+    def: >
+      コマンドのヘルプページ を開けて、
```
naoe-tatara, line 2867:
```suggestion
      [コマンドラインインターフェース](#cli)において、調べたいコマンドのヘルプページ を開く時に使う。
```

glossary.yml
```diff
@@ -2855,6 +2855,18 @@
...
+      コマンドの使い方を伝えやどう動きの詳細説明を組み込まれていのこと。
```
naoe-tatara, line 2868:
```suggestion
      コマンドの使い方や機能の詳細な説明を参照することができる。
```

glossary.yml
```diff
@@ -2855,6 +2855,18 @@
...
 
 
+- slug: help_command
+  en:
+    term: "help (command)"
+    def: >
+      Opens a help page for the command. The system provides detailed information
+      on how the command works and how it is used.
+  ja:
+    term: "ヘルプ(機能)"
```
naoe-tatara, line 2865:
```suggestion
    term: "ヘルプ(コマンド)"
```

naoe-tatara (changes requested):
I tried making it clearer that this definition is about the help command that can be used with a  command of interest in CLI. If I misunderstand, please let me know.

TomKellyGenetics:
Thanks @naoe-tatara I had the R help system in mind but I think your definition applies to the case as well.

elletjies:
Closing this pull request since the last activity took place in April 2021. Please feel free to reopen in the future to make the necessary changed.

glossary.yml
TomKellyGenetics: Addressed by ae6d55784bde84d091b4d90eee71f474ce5bfe31

TomKellyGenetics:
@elletjies The Japanese version has already been approved and proposed changes were already implemented. All that remains is a review of the English version.

elletjies:
@all-contributors please add @naoe-tatara for review,translation

Issue #311, opened 2021-05-06
[2021-05-06]: [Italian]


lucaferranti:
Hi everyone,

this project looks like fun. I'd be glad to contribute with my mother language (Italian). I noticed the language is not supported yet, should I open a PR to add the infrastructure? I guess I could figure out how to do it by reverse engineering #224, but I wanted to double check first that adding a new language is not an "admin privilege", as I didn't find instructions about it in the `CONTRIBUTING.md`.

cheers
luca

fmichonneau:
Hi Luca,

Feel free to submit a PR to add the infrastructure needed to support Italian. We'd love to have more languages supported. We are looking forward to your contributions. Would you also be willing to be an editor for the Italian language?

Thanks!

lucaferranti:
I am not sure what you mean by editor exactly, but if it means being willing to keep adding translations and review others' PR than sure count me in.

zkamvar:
Hi Luca,

You are correct that the documentation needs to be updated for adding a new language. I believe what you need to do is to create a markdown file that corresponds to your language that needs two things:

1. a YAML header that specifies the permalink relative to the root of the site (i.e. `permalink: /it/`). 
2. the main body should be `{% include glosarry.html %}` for Jekyll to insert the Italian content from the main glossary.

For example, https://github.com/carpentries/glosario/blob/master/es.md

lucaferranti:
thank you for the instructions. From what I could tell by reverse engineering #224 , I also updated the `_config.yml` and added the language code to `utils/check-glossary.py`, no idea what this last one does though :D. Anyway, I opened a PR #313, let me know if I got some settings wrong

fmichonneau:
Thanks Luca! I invited you to join the "glosario-editors-it" team. 

If you'd like, you can also join The Carpentries Slack (https://swc-slack-invite.herokuapp.com/) and there, we have a glosario channel.

PR #312, opened 2021-05-06
Update CONTRIBUTING.md

zkamvar:
I'm updating contributing to help address the confusion about contributing a new language (from #311)

Another thing that should probably be addressed in a future PR is standardizing the current language files since we have both the `layout` and `direction` yaml items, but I think only `direction` is honored.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -137,7 +137,6 @@ following to the `_config.yml` file under the `language:` entry, respecting the 
 To make sure that entries in the glossary for the language added will be recognized, you need to add the two-letter ISO code for the language in the Python script `[utils/check-glossary.py](https://github.com/carpentries/glosario/blob/master/utils/check-glossary.py#L33)`:
 
 ```python
-NTRY_OPTIONAL_KEYS = {'ref'}
 ENTRY_LANGUAGE_KEYS = {'af', 'am', 'ar', 'bn', 'de', 'en', 'es', 'fr', 'he', 'it', 'ja', 'nl', 'pt', 'zu'}
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,7 +59,7 @@ Here is an example of how your glossary entry should be structured:
 
 Once your term and definition(s) are complete, [make a Pull Request][pr-guide] back to the main repository.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -86,6 +86,63 @@ You can then fill in the `term`, `def`, and (if appropriate) `acronym` for that
       toutes les fonctions.
 ```
 
+## 3. Adding a new language
+
+If you want to add a term or define a term in a new language, you can quickly check if your language is defined in glosario by heading to `https://glosario.carpentries.org/[lang]`, replacing `[lang]` with the [two-letter ISO 639 language code][iso639-table-en]. If you see a
```
fmichonneau, line 91: Would it be easier to point to https://glosario.carpentries.org/ and check on the list of languages on the right?

fmichonneau:
Thanks for starting this Zhian!

I added some info about the other 2 files that need to be edited. Could you please review my changes?

CONTRIBUTING.md
zkamvar: Looking at the home page definitely works, but it will eventually get frustrating when the number of languages grows. I'm okay with either one.

fmichonneau: good point, let's keep the way you wrote it.

CONTRIBUTING.md
```diff
@@ -86,6 +86,62 @@ You can then fill in the `term`, `def`, and (if appropriate) `acronym` for that
       toutes les fonctions.
 ```
 
+## 3. Adding a new language
```
zkamvar, line 89: I added the numbering here because the previous sections were numbered, however, they don't need to be sequential and given the subheadings below, I think it would be best to remove the sequential numbers from the 2nd-level headings.

CONTRIBUTING.md
```diff
@@ -113,13 +113,18 @@ Each markdown file requires two things: a yaml header that defines the permalink
 ```markdown
 ---
 permalink: /it/
+layout: glossary-ltr
```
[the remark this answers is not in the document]
zkamvar: Okay, looking at this a bit more, I now understand it and will comment in the related issue.

fmichonneau (approved):

PR #314, opened 2021-05-07
update yaml direction item for it.md

zkamvar:
From my reading of glossary.html, we are using direction, not layout:

https://github.com/carpentries/glosario/blob/f977df7f2b13707799521feee93d0951fde1b4db/_includes/glossary.html#L2-L6

### DIFF
--- a/it.md
+++ b/it.md
@@ -1,5 +1,6 @@
 ---
 permalink: /it/
-layout: glossary-ltr
+direction: ltr
 ---
+
 {% include glossary.html %}


lucaferranti:
hups, sorry for that :D for what is worth, it seems to me that most of the other markdown files also use `layout` 🤔

fmichonneau:
We use both. The confusing aspect here is that `ltr` is the default and is omitted in most places, while it's required on `rtl` pages (e.g., https://github.com/carpentries/glosario/blob/master/ar.md#L4). I'll update the yaml headers to include the `direction` entry everywhere and update the docs accordingly.

PR #319, opened 2021-05-07
Standardise layouts

zkamvar:
When updating CONTRIBUTING in #312, I noticed that we had direction-specific templates (`glosario-ltr` and `glosario-rtl`) and a direction yaml item. I was confused by this arrangement: https://github.com/carpentries/glosario/pull/314#issuecomment-834575155

There were two problems with this arrangement:

1. The aforementioned confusion
2. Changes that would happen in one layout would not necessarily be propagated to the other layout. 

I've consolidated the two layouts in `_layouts/glossary.html`, updated `_includes/title.html`, and modified all the yaml items for the landing pages to use `layout: glossary`


fmichonneau (approved):

fmichonneau:
Thanks Zhian!

PR #321, opened 2021-05-19
Added term "Sanity Check" (English)

bawee:
Added the term "Sanity check" with the following description:
A basic test to see if the outcome of a calculation, script or analysis makes sense or is true.  
This can be performed by visualisation or by simply inspecting the outcome.

## Author: Bryan Wee

- 

## Language: English

- 

## Terms defined: Sanity check

- 
- 
-

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -5446,6 +5446,12 @@
       environment that is only allowed to perform a restricted set of operations for
       security reasons.
 
+- slug: sanity_check
+  en:
+    term: "sanity check"
+    def: >
+      A basic test to see if the outcome of a calculation, script or analysis makes sense or is true.  
+      This can be performed by visualisation or by simply inspecting the outcome. 
 
 - slug: scalar
   en:


fmichonneau:
Hi Brian,

Thanks for your contribution. The term you introduce is ableist language. Several projects are moving away from using such terms in their code (e.g., [Android](https://source.android.com/setup/contribute/respectful-code), [Twitter](https://twitter.com/TwitterEng/status/1278733305190342656)). I defer to the @carpentries/glosario-editors-en to address this. A possible way could be that the term gets included in the glossary but the definition should mention that it is ableist and links to alternative terms.

bawee:
Thanks François, I wasn't aware of this. Thanks for bringing it to my attention. Yes, I agree that suggesting more inclusive alternatives could be the way forward, to bring awareness to the use of such language.

jsteyn (approved):

elletjies:
@all-contributors please add @bawee for translation

PR #325, opened 2021-06-02
Translate 5 terms (Spanish)

bioinfowlad:
Added Spanish definitions for ASCII, asynchronous, bit, byte, data frame.

(change proposed as part of instructor training checkout)

## Author: 

- bioinfowlad (Wladimir Labeikovsky)

## Language: 

- Spanish (es)

## Terms defined:

- ASCII
- asynchronous
- bit
- byte
- data frame


nicoguaro (changes requested):
Can you limit the length of the lines? Now some of them are around 170 characters.

bioinfowlad (commented):
Done! (Shortened Lines)

nicoguaro (approved):

PR #330, opened 2021-06-28
Added definition "source code" (English)

fmichonneau:
contribution by @badryan

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -5810,6 +5810,13 @@
       A set of programs that are built, tested, and distributed as a collection so
       that they can run together.
 
+-slug: source_code
+  en:
+    term: "source code"
+    def: >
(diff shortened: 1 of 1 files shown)


annajiat (changes requested):
Should a space be added between - and slug?

glossary.yml
```diff
@@ -5810,6 +5810,13 @@
       A set of programs that are built, tested, and distributed as a collection so
       that they can run together.
 
+-slug: source_code
```
zkamvar, line 5813:
```suggestion
- slug: source_code
```

zkamvar: This would address the test failure and @annajiat's comment.

annajiat (approved):

PR #370, opened 2021-08-30
Add definition for independent_variable (English)

lphegley:

## Author: Lauren Phegley
(if you are asking for the author of the definition, I paraphrased from [ThoughtCo.](https://www.thoughtco.com/definition-of-independent-variable-605238))

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -3336,6 +3336,11 @@
 
 
 - slug: independent_variable
+  en:
+    term: "independent variable"
+    def: > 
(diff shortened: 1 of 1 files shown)


annajiat (commented):
its good to see that given definition explains other(s) dependent variable depends on the independent variable.

should we include the fact that the independent variable itself does not depend on another variable?

elletjies:
@all-contributors please add @lphegley  for translation

PR #377, opened 2021-09-10
Update glossary.yml

annajiat:
fixes #330
@zkamvar for some reason, the previous merge didn't fix, hence pushing again.

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -5979,9 +5979,9 @@
   en:
     term: "source code"
     def: >
-    Source code or, simply, code, is the origin of executed code (either by means of
-    an interpreter or compiler). It's the primarily human-produced series of commands
-    that make up a program. (Note: Automatic code generators exist for some applications)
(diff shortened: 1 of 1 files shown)


zkamvar (changes requested):
The problem is that lines 5982-5984 need to be indented. Yaml thinks that these are new key/value pairs and it's complaining because it can't find the values ("syntax error: could not find expected ':' (syntax)").

annajiat:
Thank you @zkamvar for the guidance, done! Hope its okay now.

zkamvar (approved):

zkamvar:
> Thank you @zkamvar for the guidance, done! Hope its okay now.

Looks good me and the checks are passing! 

In the future, be sure to wait for all the checks to pass before merging (and yes, I know it can be frustrating since the yaml linter will only report one error at a time, even if there are multiple errors).

PR #479, opened 2022-03-31
Adding German for While

kkauder:
## Author: Kolja Kauder

- 

## Language: German

- 

## Terms defined: while loop

**QUESTION: I'm using references to loop_body and true, but they probably point to the English ones if anything. Do I need to add a slug: Schleifenkörper, or add a "de" section to slug: loop_body?**

- 
- 
-


VScharf7 (changes requested):
change Kontrollstruktur to **Anweisung**

elletjies:
@all-contributors please add @kkauder for translation

PR #506, opened 2022-07-07
Changing the format

Mona-2022:

## Author: 

- 

## Language: 

- 

## Terms defined:

- 
- 
-


zkamvar:
Hi @Mona-2022,

Thank you for the contribution, but I'm going to close this PR because [level 1 headings should only be used to introduce the page as a whole](https://webaim.org/techniques/semanticstructure/#headings). The credits section should remain as level 2

PR #522, opened 2022-08-15
Add netlify and Code of Conduct links

zkamvar:
We use netlify to deploy glosario, but we need to switch to a Pro plan because we are reaching the limits of our deployment capabilities. Luckily, [Netlify has an Open Source Plan policy](https://www.netlify.com/legal/open-source-policy/), which effectively stipulates that we can have a prorated pro plan, as long as our project is 

1. open source (it is), 
2. has a Code of Conduct (it does) that is clearly visible (in this PR, which was sorely needed), 
3. it says that it's deployed by Netlify (in this PR), and 
4. must not be a commercial product (it is not)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .Rproj.user
 .jekyll-cache
 .jekyll-metadata
+.vendor/
 Glossary.egg-info
 MANIFEST
 R/sysdata.rda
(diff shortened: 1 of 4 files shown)


zkamvar:
Since this is a relatively minor change that does not affect the translations themselves (and because I don't want to waste the time of collaborators outside The Carpentries), I'm going to merge this.

Issue #531, opened 2022-10-21
2022-10-21: Build Failure


zkamvar:
I believe the token needs to be updated, but _really_ I think we need to just avoid these actions that do what we can do in the action directly. 

From https://github.com/carpentries/glosario/actions/runs/3278386716/jobs/5396715910

```
Run JamesIves/github-pages-deploy-action@4.1.5
  with:
    branch: site
    folder: _gh-site
    token: ***
    clean: true

    GitHub Pages Deploy Action 🚀

    🚀 Getting Started Guide: https://github.com/marketplace/actions/deploy-to-github-pages
    ❓ Discussions / Q&A: https://github.com/JamesIves/github-pages-deploy-action/discussions
    🔧 Report a Bug: https://github.com/JamesIves/github-pages-deploy-action/issues

    📣 Maintained by James Ives: https://jamesiv.es
    💖 Support: https://github.com/sponsors/JamesIves
Checking configuration and starting deployment… 🚦
Deploying using Deploy Token… 🔑
Configuring git…
/usr/bin/git config user.name nicoguaro
/usr/bin/git config user.email nicoguarin@gmail.com
/usr/bin/git config --local --unset-all http.https://github.com/.extraheader
Unable to unset previous git config authentication as it may not exist, continuing…
/usr/bin/git remote rm origin
/usr/bin/git remote add origin ***github.com/carpentries/glosario.git
Git configured… 🔧
Starting to commit changes…
/usr/bin/git ls-remote --heads ***github.com/carpentries/glosario.git refs/heads/site
30a93cb8d95f7c638e0606e4[2](https://github.com/carpentries/glosario/actions/runs/3278386716/jobs/5396715910#step:6:2)ea2d[3](https://github.com/carpentries/glosario/actions/runs/3278386716/jobs/5396715910#step:6:3)[4](https://github.com/carpentries/glosario/actions/runs/3278386716/jobs/5396715910#step:6:4)062edb88[5](https://github.com/carpentries/glosario/actions/runs/3278386716/jobs/5396715910#step:6:5)	refs/heads/site
Creating worktree…
/usr/bin/git fetch --no-recurse-submodules --depth=1 origin site
From https://github.com/carpentries/glosario
 * branch            site       -> FETCH_HEAD
 * [new branch]      site       -> origin/site
/usr/bin/git worktree add --no-checkout --detach github-pages-deploy-action-temp-deployment-folder
[...]

PR #532, opened 2022-10-21
[workflows] test permissions bump

zkamvar:
This will address #531 by updating the permissions in the workflow.


zkamvar:
The builds are working now an I am not sure why, so I'm going to close this.

PR #541, opened 2022-11-29
Added Kiswahili translations for ‘condition’

NOngeso:

## Author: 

- NOngeso

## Language: 

- sw

## Terms defined:

- condition
- 
-

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -2625,7 +2625,7 @@
       'n Fout of ander onverwagte gebeurtenis wat die normale vloei van beheer ontwrig.
       
   sw:
-    term:"hali"
+    term: "hali"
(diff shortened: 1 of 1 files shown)


glossary.yml
```diff
@@ -2624,7 +2624,10 @@
     def: >
       'n Fout of ander onverwagte gebeurtenis wat die normale vloei van beheer ontwrig.
       
-      
+  sw:
+    term:"hali"
```
elletjies, line 2628: I think there must be a space between the : and " on line 2628

elletjies (changes requested):

elletjies (approved):

PR #555, opened 2022-12-12
Corrected typo :+ to +

elletjies:

## Author: 

- 

## Language: 

- 

## Terms defined:

- 
- 
-


jsteyn (changes requested):
It's is still missing the open quotes. Should be "+1"

jsteyn (approved):

PR #557, opened 2022-12-14
Created Southern Sotho md file

elletjies:
@zkamvar I noticed there are not st.md file so created it. We do have Southern Sotho on the main page here (https://glosario.carpentries.org) - will this fix the issue?

### DIFF
--- a/st.md
+++ b/st.md
@@ -1,0 +1,6 @@
+---
+permalink: /st/
+layout: glossary
+direction: ltr
+---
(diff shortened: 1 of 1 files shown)


zkamvar (approved):

zkamvar:
I think that should fix it! Let's see what happens :)

elletjies:
@zkamvar it worked! Thanks!

PR #571, opened 2023-02-20
Update Portuguese definition of 'argument' and add 'parameter'

ianvcaldas:
Updated Portuguese definition of "argument" to match the content of the definition in other languages. That definition compares arguments to parameters; a Portuguese definition of "parameter" didn't exist, so I also added it.

## Author: 

- Ian V. Caldas

## Language: 

- Portuguese

## Terms defined:

- Argument
- Parameter

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -983,8 +983,9 @@
   pt:
     term: "argumento"
     def: >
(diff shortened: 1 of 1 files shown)


elletjies:
@all-contributors please add @ianvcaldas for translation

PR #590, opened 2023-06-29
Editing special characters in Bayes' Theorem definition

jcostacurta11:
Bayes' Theorem definition uses yaml special character |, which was causing weird formatting (column breaks) on the glosario website. I've attempted to fix this by adding ` quotes around each instance of the character, but I'm not sure how to check if it solves the issue.


elletjies:
@all-contributors please add @jcostacurta11 for translation

Issue #576, opened 2023-03-08
2023-03-08: English/Deep learning terms


dsmits:
I am currently working on the glossary for the [deep learning lesson](https://github.com/carpentries-incubator/deep-learning-intro).

There are a couple of terms that I need that are currently missing from glosario:

- artificial intelligence
- convolutional neural network (CNN)
- recurrent neural network (RNN)
- epoch
- learning rate
- class imbalance
- hidden layers

dsmits:
I can write these definitions myself, just let me know if it's ok if I add these.

elletjies:
@dsmits please go ahead and submit the terms :)

svenvanderburg:
@elletjies #584 is merged, but I don't see the new terms appearing in glosario. For example: https://glosario.carpentries.org/en/#cnn

I see in the Github Action that Deploy was skipped: https://github.com/carpentries/glosario/actions/runs/5875720776/job/16019059004 

How do we get this to appear in the glosario?

elletjies:
@svenvanderburg, it seems to appear in the glossary.yml file but not on https://glosario.carpentries.org. @zkamvar will you be able to assist?

zkamvar:
I've opened https://github.com/carpentries/glosario/pull/596 to fix the issue and I've tested it on my own fork and confirm that it works.

elletjies:
Thank you for fixing this @zkamvar!

svenvanderburg:
All the terms added in this issue still do not show up, for example this term is still not in there: https://glosario.carpentries.org/en/#cnn Could you reopen this issue? @elletjies @zkamvar

svenvanderburg:
It seems like the yaml was not formatted correctly, I think this fixes it: https://github.com/carpentries/glosario/pull/598

svenvanderburg:
Yes, it is fixed now! Thanks a lot @elletjies and @zkamvar :pray:

elletjies:
@all-contributors please add @dsmits  question

elletjies:
@all-contributors please add @svenvanderburg  question

PR #596, opened 2023-09-01
Update GitHub Workflows

zkamvar:
## Problem

As noted in https://github.com/carpentries/glosario/issues/576#issuecomment-1702659711

The github pages workflow has not been running since August. After some inspection, this was due to a change in behaviour where the runner token was not detected during the build sometime after August 1.

## Solution

**please squash merge** this PR

I have fixed the pressing issue by adding a `token: ${{ secrets.GITHUB_TOKEN }}` to the `gh-pages-deploy-action`, which is automatically scoped to the type of run so that `pull_requests` do not have any write permissions (see [the automatic token documentation](https://docs.github.com/en/actions/security-guides/automatic-token-authentication)). 

Here are the other changes I've implemented:

1. added the `workflow_dispatch` button to the build and deploy workflow so we can rebuild on the fly without adding a fake commit
2. update many of the actions to `@v3` so that we no longer get warnings about deprecated node versions
3. switch the JamesIves action to `@v4` so that we can benefit from security updates in that branch

### DIFF
--- a/.github/workflows/build-and-deploy.yaml
+++ b/.github/workflows/build-and-deploy.yaml
@@ -1,4 +1,4 @@
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 
 name: Build and Deploy Glossary Website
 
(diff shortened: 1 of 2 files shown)


zkamvar:
I'm going to merge this without review because it _is_ important for deployment and I've gotten implicit review from @elletjies.

PR #630, opened 2024-03-05
Added French translation for Accuracy

elletjies:
## Author: 

- @emdotson23

## Language: 

- French

## Terms defined:

- Accuracy
- 
-

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -455,9 +455,18 @@
   en:
     term: "accuracy"
     def: >
-      Statistical measure of a classification model which gives the proportion 
-      of correct predictions among total number of cases. It is calculated as 
-      Accuracy = (TP+TN)/(TP+TN+FP+FN)
+      A statistical measure of a classification model which gives the proportion 
+      of correct predictions among a total number of cases. It is calculated as 
+      Accuracy = (TP+TN)/(TP+TN+FP+FN). 
+      
+      - TP = True Positive
+      - FP = False Positive
+      - FN = False Negative
+      - TN = True Negative
+
+      Accuracy and precision are both ways to measure results. Accuracy measures how close results are 
+      to the true or known value. Precision, on the other hand, measures how close results are to one another. 
+      
   es:
     term: "precisión"
     def: >
@@ -483,7 +492,9 @@
   fr:
     term: "Précision"
     def: >
-      l'exactitude est la proximité des mesures à une valeur spécifique, tandis que la précision est la proximité des mesures les unes par rapport aux autres. Il est calculé comme Exactitude = (TP+TN)/(TP+TN+FP+FN)
+      L'exactitude et la précision sont deux façons de mesurer les résultats. 
+      L'exactitude mesure à quel point les résultats sont proches de la valeur vraie ou connue. 
+      La précision, quant à elle, mesure la proximité des résultats les uns par rapport aux autres. 
 
 - slug: actual_result
   de:
--- a/glossary.yml
+++ b/glossary.yml
@@ -480,6 +480,10 @@
     def: >
       Dit is 'n statistiese maatstaf van 'n klassifikasiemodel wat die proporsie van korrekte voorspellings gee onder die totale aantal gevalle. 
       Dit word bereken as Akkuraatheid = (TP+TN)/(TP+TN+FP+FN).
+  fr:
+    term: "Précision"
+    def: >
+      Mesure statistique d'un modèle de classification qui donne la proportion de prédictions correctes parmi le nombre total de cas. Il est calculé comme Exactitude = (TP+TN)/(TP+TN+FP+FN)
 
 - slug: actual_result
   de:
--- a/glossary.yml
+++ b/glossary.yml
@@ -483,7 +483,7 @@
   fr:
     term: "Précision"
     def: >
-      Mesure statistique d'un modèle de classification qui donne la proportion de prédictions correctes parmi le nombre total de cas. Il est calculé comme Exactitude = (TP+TN)/(TP+TN+FP+FN)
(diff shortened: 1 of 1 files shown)


jsgro (changes requested):
Je pense que mettre "Exactitude" devant la formule est en contradiction avec les définitions de [fr.wikipedia.org/wiki/Exactitude_et_précision](https://fr.wikipedia.org/wiki/Exactitude_et_précision) qui suggèrent que ce sont 2 entités différentes:
"l'exactitude est la proximité des mesures à une valeur spécifique, tandis que la précision est la proximité des mesures les unes par rapport aux autres."

jsgro:
Hello again... I am new to this Glossary thing... It think it is a great idea.

This is a very difficult translation as "exactitude" and "precision" are often used interchangeably.     
I don't speak German, but I used their translation [Genauigkeit](https://translate.google.com/?sl=auto&tl=en&text=Genauigkeit&op=translate) in Google Translate and both "exactitude" and "precision" are suggested translations amongst more words.

This web https://asana.com/resources/accuracy-vs-precision has a good explanation of both terms.

```
Accuracy evaluates how close you are to the true value of the measurement, 
while precision shows how close the measured values are to one another. 
```
The (original language?) English is about *accuracy* and **not about** *precision*. 

AT FIRST I thought that using *exactitude* was not the correct word for the French translation because *precision* was used in the other languages that I can understand (*e.g.* Spanish.) However, the above definition as `Accuracy evaluates how close you are to the true value of the measurement` suggests that SOME of the other language translations are not reflecting the correct translation of the English `accuracy`.

Since "exactitude" is a feminine word in French, your definition should be slightly altered when providing the formula from "Il" to "Elle" as and adding the feminine "e" ending to the calculation word calculé***e***.

 l'exactitude est la proximité des mesures à une valeur spécifique, tandis que la précision est la proximité des mesures les unes par rapport aux autres. ***Elle*** est calculé***e*** comme `Exactitude = (TP+TN)/(TP+TN+FP+FN)`

This French definition therefore does not contain a translation of the English, as used by the other langages, but does contain the contrast definition between the 2 words, which is useful.

elletjies:
Thank you @jsgro! Can you help me with a full definition for this translation (as you suggested above). I can then add the changes in GitHub?

jsgro:
Greetings Angelique. I see that you are the author of the page at https://carpentries.org/blog/2024/02/Glosario-Codefest/ and there will be a "Glosario Codefest" tomorrow (May 15) it seems, and the Univ. of Wisconsin-Madison (USA) will participate. Trying to understand how the whole thing works I stumbled on your earlier review and thought that this would be a rather easy definition. But in fact it's not!

The first problem is that the English version is definitely not the best, as it is not a broad definition but a definition based on "statistical models" and related to what is called a "Confusion Matrix" as it is explained in this post: [Confusion Matrix to No Confusion Matrix in Just 5mins](https://pub.towardsai.net/confusion-matrix-179b9c758b55)
and also present the following formula 

`Accuracy = (TP+TN)/(TP+FP+FN+TN)` but the [Carpentry glossary](https://github.com/carpentries/glosario/blob/main/glossary.yml) definition of "Accuracy" does not specify what these letters mean:

- TP = True Positive
- FP = False Positive
- FN = False Negative
- TN = True Negative

As these terms are NOT defined in the original English proposition, the meaning of the formula becomes moot and senseless! and are unfortunately blindly copied in all other language translations.

Since the English definition is *biased* on the general idea of a "Statistical model," all other languages have simply translated the same bias... I think it's not the best approach for this particular word. This *bias* of definition is likely due to the fact that some author of a lesson wanted to define *Accuracy* for that particular lesson, but it is not a general definition.

Since we can only alter the French defintion, I suggest that we make use of the simpler definition I mentioned previously on this web page:  https://asana.com/resources/accuracy-vs-precision 

```
Accuracy and precision are both ways to measure results. Accuracy measures how close results are 
to the true or known value. Precision, on the other hand, measures how close results are to one another. 
```

Which translate in French as follows, and we can also add: *Accuracy = Exactitude* to clarify the specific translation of that word:


```
Accuracy = Exactitude
L'exactitude et la précision sont deux façons de mesurer les résultats. 
L'exactitude mesure à quel point les résultats sont proches de la valeur vraie ou connue. 
La précision, quant à elle, mesure la proximité des résultats les uns par rapport aux autres. 
```

The advantage of this definition is:

1. It is very clear, without jargon or formula
2. It offers a specific definition for both accuracy and precision which are often used interchangeably but are not the same.
3. It is not dependent on any "Model" or "Statistical Model"

I would even strongly suggest that this should be the basis of the definition of *Accuracy* in all other languages.

Best wishes,
JYS

elletjies:
@jsgro please feel free to merge the PR

jsgro:
> **@jsgro thank you for these suggestions! I have added the language you suggested. Looking forward to seeing you at the Glosario event!**

Sorry to have missed the event, got the time wrong... I made the mistake to rely on the "[Add to your calendar](https://calendar.google.com/calendar/event?action=TEMPLATE&tmeid=MTF0djFwbW8zdGY4OTlsZHEyNDlnNzlnZnUgb3NldXVvaHQwdHZqYm9rZ2czbm9oOGM0N2dAZw&tmsrc=oseuuoht0tvjbokgg3noh8c47g%40group.calendar.google.com)" link which give the South African time... i.e. 5PM while here it would have been 9AM... 

Glad you were able to add the suggestions.

> **@jsgro please feel free to merge the PR**

I cannot find a button that would say "approve" ... Any idea where that is?

elletjies:
@jsgro you should see a review button at the top right of the page. You can click on that and then say approve.

jsgro (approved):
Good job!

froggleston (approved):
LGTM! 👍

PR #660, opened 2024-06-06
Translated "Unicode" to Ukrainian

Dod900ls1:

## Author: 

- Yehor Boiar

## Language: 

- Ukrainian

## Terms defined:

- Unicode

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -9961,6 +9961,14 @@
       Une norme qui définit des codes numériques pour plusieurs milliers de caractères 
       et de symboles. Unicode ne définit pas comment ces nombres sont stockés ; cela est 
       fait par des normes comme [UTF-8](#utf_8).
+  uk:
+    term: "Unicode"
+    def: >
+      Промисловий стандарт і специфікація, розроблені для забезпечення цифрового представлення 
+      символів усіх писемностей світу та спеціальних символів. Unicode надає кожному символу
+      його унікальний код. Специфікації Unicode постійно переглядаються й оновлюються для 
+      додавання нових мов і символів. Стандарт не визначає, як ці числові коди зберігаються
+      в пам'яті комп'ютера; для цього існують інші стандарти збереження, такі як [UTF-8](#utf_8).
 
 - slug: unit_test
   ref:


glossary.yml
```diff
@@ -9961,6 +9961,14 @@
...
       et de symboles. Unicode ne définit pas comment ces nombres sont stockés ; cela est 
       fait par des normes comme [UTF-8](#utf_8).
+  uk:
+    term: "Unicode"
+    def: >
+      Промисловий стандарт і специфікація, розроблені для забезпечення цифрового представлення 
+      символів усіх писемностей світу та спеціальних символів. Unicode надає кожному символу
+      унікальний код. Специфікації Unicode постійно переглядаються й оновлюються для 
+      додавання нових мов і символів. Стандарт не визначає, як ці числові коди зберігаються
+      в пам'яті комп'ютера; для цього існують стандарти збереження, такі як [UTF-8](#utf_8).
```
Herobread, line 9971: Although it's better than the English version, this doesn't match it. Would this be better?

`Стандарт, який визначає числові коди для багатьох тисяч символів і знаків. Unicode не визначає, як ці числа зберігаються; це робиться за допомогою таких стандартів, як [UTF-8](#utf_8).`

Herobread (changes requested):

glossary.yml
Dod900ls1: I don't think so. I think even though my version is a little bit longer, it contains much more insight into the term. In particular, I don't like those parts:

- "коди для багатьох тисяч символів і знаків" - It doesn't tell what symbols it assigns the code to, and it doesn't mention that this code has to be unique.
- It doesn't tell that this standard is constantly updating, unlike, for example, the ASCII or UTF-8 standards, which is essential.

However, I do think that it would be the best if it was shorter. I'm open to any suggestions in this direction!

olexandr-konovalov, line 9971: "існують інші стандарти збереження" звучить краще, якщо читати вголос

glossary.yml
```diff
@@ -9961,6 +9961,14 @@
       Une norme qui définit des codes numériques pour plusieurs milliers de caractères 
       et de symboles. Unicode ne définit pas comment ces nombres sont stockés ; cela est 
       fait par des normes comme [UTF-8](#utf_8).
+  uk:
+    term: "Unicode"
+    def: >
+      Промисловий стандарт і специфікація, розроблені для забезпечення цифрового представлення 
+      символів усіх писемностей світу та спеціальних символів. Unicode надає кожному символу
+      унікальний код. Специфікації Unicode постійно переглядаються й оновлюються для
```
olexandr-konovalov, line 9969: мабуть "свій унікальний" або "його унікальний" ?

glossary.yml
olexandr-konovalov: Good  question, leading to a conceptual discussion (and possibly to feeding back @Dod900ls1's additions to the English version). English version is usually a good start. but we do not have to do a word-by-word translation. If the exact translation is good. then it's fine. but if you find that it does not express the concept well enough, or that the concept is better explained in Ukrainian source you have used in other formulations, consider other options too, trying to follow the concise style of Glosario.

In this particular example, @Dod900ls1's text is still short enough, and adds useful context, helpful for understanding.

olexandr-konovalov (approved):

olexandr-konovalov:
LGTM! @Herobread to you too?

Herobread (approved):

olexandr-konovalov (approved):
@elletjies could you please merge this! thanks!

PR #698, opened 2024-06-11
Fix links in term GPU

Herobread:
updated brackets, so that they create link
image


olexandr-konovalov (approved):
Well done. @elletjies this fixes both English and Ukrainian versions - please merge

Issue #254, opened 2020-12-15
2020-12-15: Japanese glosario page is not alphabetical order


masamiy:
https://carpentries.github.io/glosario/ja/ lists Japanese entries based on the first character of the entry. It means that the entries are not categorised by Japanese alphabet (nor English alphabet), but characters. The last entry, 'function', should be top of the current list as it is read as 'kansuu' if terms are categorise by Japanese alphabet. 
As there are 46+ characters in Japanese alphabet, I feel we need to have some indexing strategy.

baileythegreen:
@masamiy  

The order of the entries is determined by a `sort` function on line 16 of `_includes/glossary.html`, which operates on individual characters. It may be that for languages such as Japanese we need to find a different solution entirely. The `sort` function currently being used is a liquid one, and I very much doubt they have a different one that will sort Japanese correctly. I am familiar with the website infrastructure and the code, but I don't know the Japanese alphabet, so this isn't something I can fix on my own.

I can see two options for solving it:

1. We find or write a function in something like Ruby or Python to sort Japanese (and any other language that has this problem), based on an input list of the alphabet, if need be.
1. We move to a different system for storing definitions other than a YAML file so that the sorting can take place at a slightly different step. An example would be an SQLite database which can export its contents, or part of its contents, as a YAML or other config-type file. This involves more changes to the infrastructure, though.

Perhaps @fmichonneau or @gvwilson will have another idea?

fmichonneau:
It looks like option 1 is going to be the way to go.
From a quick search, I saw [mecab](https://taku910.github.io/mecab/) being mentioned regularly but that's Japanese-specific and wouldn't work for Arabic, Hebrew, Amharic, etc.
From my limited understanding of this, I think the [ICU library](http://site.icu-project.org/design/cpp) would order the characters correctly. In R, it's implemented by the `stringi`/`stringr` packages, in Python, by [PyICU](https://pypi.org/project/PyICU/).

baileythegreen:
I think option 1 is certainly easier to implement in the short-term. I can take a stab at writing Python code to do this, though I may need someone to verify the output in those languages.

If I do this, unless someone has an objection, I'll probably try to remove the `sort` logic from `_includes/glossary.html` entirely and use one script to do all alphabetising, rather than have it happen in different places based on the language in question.

masamiy:
[...]

PR #748, opened 2024-09-17
Add support for externalised sorting per language

### LINKED ISSUE #724: [2024-06-24]: [Alphabetical ordering for Ukrainian and Portuguese languages]
The right navigation bar with letters at https://glosario.carpentries.org/uk/ does not follow the order of letters in Ukrainian alphabet, placing  "І" at the last place. Is there a way to enforce the fixed order of the letters?

 
Screenshot 2024-06-24 at 18 53 38

froggleston:
This PR attempts to fix #724 and #254 by moving away from liquid's in-built `sort` function and to use a bespoke sorting provided by `sort-glossary.py`.

There are two important changes:

- The makefile now runs an external python script to sort the alphabet of each language. As each language does not have the same natural sorting of terms, this needs to be stored per language (see below)
- The main `glossary.yml` file used to be copied into the `_data` folder to act as the served content to liquid/jekyll. This has now changed in that the python script creates a `_data/{lang}/glossary.yml` file for each language and the template loads the specific data file based on the selected language. It also dumps out non-Latin characters in Unicode to avoid ambiguity in rendering - humans are not meant to read this `_data` version of the glossary!


utils/sort-glossary.py
```diff
@@ -0,0 +1,335 @@
...
+    ('tr', 'Turkish'),
+    ('tw', 'Twi'),
+    ('ug', 'Uighur; Uyghur'),
+    ('uk', 'Ukrainian'),
+    ('ur', 'Urdu'),
+    ('uz', 'Uzbek'),
+    ('ve', 'Venda'),
+    ('vi', 'Vietnamese'),
+    ('vo', 'Volapük'),
+    ('cy', 'Welsh'),
```
olexandr-konovalov, line 205: @froggleston Welsh is already present above in line 47. Maybe sort items so that we can easier spot accidental duplicates?

froggleston: Great spot! I've removed the duplicates.

froggleston:
Sadly we can't add netlify builds as netlify doesn't allow system level dependencies (e.g. apt-get) at build time and we now require pyicu to be built which requires the libicu-dev package.

elletjies (approved):

PR #755, opened 2024-11-13
Translated terms to Afrikaans

elletjies:

## Author: 

- @elletjies 

## Language: 

- Afrikaans

## Terms defined:

- 
- 
-

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -9215,6 +9215,11 @@
     def: >
       Місце, де [система контролю версій](#version_control_system) зберігає файли,
       що складають проект, та метадані, які описують їхню історію.
+  af:
+    term: "bewaarplek"
+    def: >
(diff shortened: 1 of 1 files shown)


elletjies:
@froggleston I have been looking at the errors, but nothing makes sense. I have tried recreating the [PR](https://github.com/carpentries/glosario/pull/759), but the errors still exist. Could you perhaps steer me into the right direction to get the errors solved?

froggleston:
This is failing due to this problem: https://github.com/carpentries/glosario/actions/runs/11816918260/job/32921242643#step:6:15

In `browser_cache`, there is a `def` entry when this should be a language two-letter code.

Also, links to other terms need to be in the slug format and not the English term. In the example here, `#standard deviation` needs to be `#standard_deviation`: https://github.com/carpentries/glosario/pull/755/commits/20aae7d14e20d68d52c9531dfbe3a078c0b764cf#diff-f13138bcde99805cc217314b608006094ad9d22f63e79a2f3cf019acc4a07003R9350

PR #1021, opened 2025-05-28
Add Italian translation for accuracy

elletjies:
## Author: 

- @fgazzelloni

## Language: 

- Italian

## Terms defined:

- accuracy
- 
-

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -139,13 +139,6 @@
     def: >
       The difference between the measured or inferred value of a quantity and its actual value, 
       sometimes with the absolute value taken. 
-  it:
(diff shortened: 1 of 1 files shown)


glossary.yml
```diff
@@ -139,13 +139,6 @@
     def: >
       The difference between the measured or inferred value of a quantity and its actual value, 
       sometimes with the absolute value taken. 
-  it:
```
lucaferranti, line 142: this PR seems to only remove an entry instead of adding a new one. Was this a git accident?

lucaferranti (changes requested):

PR #1026, opened 2025-05-28
Translated Call Stack into French

elletjies:
## Author: 

- @TinoCoda

## Language: 

- French

## Terms defined:

- call stack
- 
-

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -2571,7 +2571,10 @@
     term: "call stack"
     def: >
       Sebuah struktur data yang berguna untuk menyimpan informasi mengenai subrutin aktif yang sedang dijalankan.
-
+  fr:
+    term: "pile d'appels"
+    def: >
+      aussi appelée pile d'exécution, il s'agit d'une structure de donnée qui enregistre des informations sur les sous routines qui sont actuellement exécutées. 
   am:
     term: የጥሪ ቁልል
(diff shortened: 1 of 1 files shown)


elletjies:
@all-contributors please add @TinoCoda for translation

glossary.yml
```diff
@@ -2571,7 +2571,10 @@
     term: "call stack"
     def: >
       Sebuah struktur data yang berguna untuk menyimpan informasi mengenai subrutin aktif yang sedang dijalankan.
-
+  fr:
+    term: "pile d'appels"
+    def: >
+      aussi appelée pile d'exécution, il s'agit d'une structure de donnée qui enregistre des informations sur les sous routines qui sont actuellement exécutées.
```
MarieHouillon, line 2577: - I would maybe switch "pile d'exécution" and "pile d'appels", as the first one seems more commonly used to me (it's also the term chosen on Wikipedia: https://fr.wikipedia.org/wiki/Pile_d%27ex%C3%A9cution).

MarieHouillon, line 2577:
```suggestion
    term: "pile d'exécution"
    def: >
      aussi appelée pile d'appels, il s'agit d'une structure de données qui enregistre des informations sur les sous-routines qui sont actuellement exécutées. 
```

MarieHouillon (commented):
Thank you for your ocntribution, here are some suggestions.

PR #1049, opened 2025-06-30
Fix broken ref in openrefine entry

yexpp:
Fixed a broken reference in the openrefine entry by changing "tidy data" to "tidy_data" in the ref field to fix the internal link.

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -8276,7 +8276,7 @@
      
 - slug: openrefine
   ref:
-    - tidy data
+    - tidy_data
   en:
     term: "OpenRefine"
     def: >


olexandr-konovalov:
Thanks @yexpp for contributing this fix. The change passes the CI. Would you like to add the fixes for two more `ref`s that were detected with your code?

Please note that in the commit message

    Fix ref slug: change 'tidy data' to 'tidy_data' in glossary.yml

part of it is redundant, since one can see what changed and in which file using `git diff`. Also, you did not fix the ref slug, you have fixed `ref` in the "openrefine" slug. Maybe try next time something in the style of

    Fix ref in the openrefine slug, broken because of missing undescores

elletjies:
@all-contributors please add @yexpp for review,question

PR #1086, opened 2026-01-15
Adding "precision" in english

bnwolford:
## Author: 

- Brooke Wolford

## Language: 

- English

## Terms defined:

- precision

### DIFF
--- a/glossary.yml
+++ b/glossary.yml
@@ -8783,6 +8783,12 @@
     def: >
       A naming style that separates the parts of a name with underscores, as in `first_second_third`.
 
+- slug: precision
+  en:
+    term: "precision"
+    def: >
+      A proportional measure of how often a model correctly identifies positive instances (e.g. true positives).
+
 - slug: preamble
   ref:
     - latex
--- a/glossary.yml
+++ b/glossary.yml
@@ -8787,7 +8787,18 @@
   en:
     term: "precision"
     def: >
-      A proportional measure of how often a model correctly identifies positive instances (e.g. true positives).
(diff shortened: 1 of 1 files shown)


sl-stlzoo (changes requested):
I would suggest clarifying this a bit from what's currently proposed. As currently written, it's a bit ambiguous that precision is scoped specifically to _the model's_ predictions, rather than _all_ (actual) positives, for example (which is closer to _recall_). 

A potential update for consideration: "A proportional measure of how often a model's positive predictions are correct (e.g., true positives)."

elletjies:
@all-contributors please add @bnwolford for translation

jen-reeve (commented):
Given that the [accuracy](https://glosario.carpentries.org/en/#accuracy) entry mentions its relationship with precision, I'd include the inverse here. And as said in a previous review, precision isn't about true positives, but the consistency of the model. And add a link to this new entry for accuracy's mention of precision

elletjies:
@all-contributors please add @jen-reeve for review

gvwilson:
1. Agree with @sl-stlzoo 's comment (first in review).
2. Agree that cross-ref to accuracy should be added.

Thanks @bnwolford for writing this - happy to review again once amended.

bnwolford:
I tried to update the definition in accordance with your suggestions.

https://github.com/carpentries/glosario/pull/1086/changes/394d2eb5186a20a4da22dc0ca8f8ff789807f12b

Thanks! @sl-stlzoo @gvwilson @jen-reeve

gvwilson:
thanks @bnwolford - this looks good to me. :+1: to merge.

elletjies (approved):

